PR #1, opened 2025-06-19
Update README_zh.md

khazic:
An extra "not supported" was written

### DIFF
--- a/README_zh.md
+++ b/README_zh.md
@@ -56,7 +56,7 @@ pip install -e .
 
 ### Checkpoint 格式转换
 
-由于 slime 使用 megatron，而 megatron 不支持不支持加载 huggingface checkpoint，我们需要将模型转换至 megatron 可以支持的 torch_dist 格式。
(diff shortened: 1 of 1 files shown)


jhinpan:
## Review Comments Addressed ✅

Thank you for the review\! I've addressed all the comments:

### 1. **Config/Constants for pass@k and reward logic**
- Created `kernelbench_config.py` with centralized configuration
- All thresholds and rewards are now configurable
- Both `data.py` and `kernel_generator.py` now use these configs

### 2. **Complete Triton operations list**
- Created `triton_ops.py` with comprehensive list based on Triton documentation
- Includes all core operations: memory, arithmetic, math, linear algebra, atomics, etc.
- Now validates against the complete set of ~50+ Triton operations

### 3. **Version compatibility with hasattr**
- Already using `hasattr` in `data.py` for checking `rollout_task_type`
- Added try/except import for config to ensure backward compatibility

### 4. **Torch operations in kernels**
- Made this configurable via `KERNELBENCH_VALIDATION["allow_torch_in_kernel"]`
- By default still restricts torch ops inside `@triton.jit` functions
- Can be enabled if needed for specific use cases

The validation is now more flexible and configurable while maintaining the core goal of preventing reward hacking.

PR #2, opened 2025-06-19
Implement partial rollout feature for SGLang

guapisolo:
### Design overview

- Support partial rollout feature. After rollout engine collects enough samples, it will send abort request to SGLang router.
- ...
[TODO]


scripts/run-qwen-1.5B.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
zhuzilin: please rename the file as `run-qwen2.5-1.5B.sh` or `run-deepseek-r1-distill-qwen-1.5B.sh`

slime/ray/buffer.py
```diff
@@ -157,20 +158,66 @@ async def get_samples(self, num_samples) -> list[Sample]:
...
-
-        assert len(samples) == num_samples
         return samples
 
     async def _get_samples_from_buffer(self, num_samples) -> list[Sample]:
         if len(self.buffer) == 0 or num_samples == 0:
             return []
-        samples = self.buffer_filter(self.buffer, num_samples)
+        
+        if self.args.partial_rollout:
```
zhuzilin: I don't thinks we should hardcode how we select the samples from the existing partial rollout functions. Can we extract these into a separate function so that user can still customize the buffer selection?

slime/rollout/rm_hub/deepscaler.py
```diff
@@ -7,7 +7,7 @@ def get_deepscaler_rule_based_reward(response, label):
     elif "###Response" in response:
         model_solution = response.split("###Response")[1]
     else:
-        return 0
+        model_solution = response
```
zhuzilin: please don't change this. this contains the format reward function for deepscaler. And I thinks for deepseek distilled qwen 1.5B, checking whether the response has `</think>` still works.

slime/ray/buffer.py
```diff
@@ -126,11 +126,12 @@ async def get_samples(self, num_samples) -> list[Sample]:
 
         samples = await self._get_samples_from_buffer(num_samples)
         num_samples -= len(samples)
+        if num_samples == 0:
+            return samples
+
```
zhuzilin: I don't think we need to move this early return?

slime/utils/arguments.py
```diff
@@ -187,45 +187,48 @@ def add_rollout_arguments(parser):
...
+                    "This parameter is used with the over_sampling_filter_input_size. "
+                    "The over sampling filter is applied only after enough data has been generated."
+                    "You could use `slime.rollout.filter_hub.over_sampling_filters.sort_by_reward_std` as an example."
                 ),
             )
-            # dynamic sampling
+            # diversity sampling
             parser.add_argument(
-                "--dynamic-sampling-filter-path",
+                "--diversity-sampling-filter-path",
```
zhuzilin: hmm... It seems that DAPO stands for Dynamic Sampling Policy Optimization, I think we'd better stay with the dynamic sampling as it is more aligned with the origin name.

slime/rollout/sglang_example.py
```diff
@@ -18,15 +19,29 @@
...
+    input_truncated_samples: int = 0
+    output_aborted_samples: int = 0
+    output_completed_samples: int = 0
+    output_truncated_samples: int = 0
+    cached_pending_samples: int = 0
+    cached_aborted_samples: int = 0
+    cached_completed_samples: int = 0
+    cached_truncated_samples: int = 0
+    diversity_filter_excluded_samples: int = 0
+    over_sampling_filter_excluded_samples: int = 0
```
zhuzilin: I'm not sure why we need to collect all these stats, and it seems a bit messy... Maybe add a `stats: dict[str, Any]` field to manage them if collecting them is necessary.

slime/rollout/sglang_example.py
```diff
@@ -37,41 +52,81 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
...
-    while True:
+
+    max_retries = 5
+    retry_count = 0
+    while retry_count < max_retries:
         try:
             async with SEMAPHORE:
                 output = await post(url, payload, use_http2=args.use_http2)
+                # print(f"Prompt: {input_text[:100]}...", flush=True)
+                # print(f"Output: {output['text'][:200]}...", flush=True)
```
zhuzilin: please remove the debugging prints.

slime/rollout/sglang_example.py
```diff
@@ -37,41 +52,81 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
...
     payload = {
-        "text": sample.prompt,
+        "text": input_text,
         "sampling_params": sampling_params,
     }
-    while True:
+
+    max_retries = 5
+    retry_count = 0
+    while retry_count < max_retries:
```
zhuzilin, line 77: the reason for having a retry is that some time sglang will be inavailable for a moment. so we don't need to add a max_retries here. If the worker died, we should stop the training.

slime/rollout/sglang_example.py
```diff
@@ -37,41 +52,81 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
...
-    sample.aborted = output["meta_info"]["finish_reason"]["type"] == "abort"
+    if "rollout_id" not in sample.metadata:
+        sample.metadata["rollout_id"] = rollout_id
+    match output["meta_info"]["finish_reason"]["type"]:
+        case "length":
+            sample.status = Sample.Status.TRUNCATED
+        case "abort":
+            sample.status = Sample.Status.ABORTED
+        case "stop":
+            sample.status = Sample.Status.COMPLETED
```
zhuzilin, line 109: maybe changing the `sample.status` into `sample.finish_reason` and do:
```python
sample.finish_reason = Sample.FinishReason[
    output["meta_info"]["finish_reason"]["type"].upper()
]
```
would be cleaner.

slime/rollout/sglang_example.py
```diff
@@ -37,41 +52,81 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
...
+        sample.response = output["text"]
 
+    response_token_ids = TOKENIZER(sample.response, add_special_tokens=False)["input_ids"]
     sample.tokens = prompt_tokens_ids + response_token_ids
     sample.response_length = len(response_token_ids)
-    sample.truncated = output["meta_info"]["finish_reason"]["type"] == "length"
-    sample.response = output["text"]
-    sample.aborted = output["meta_info"]["finish_reason"]["type"] == "abort"
+    if "rollout_id" not in sample.metadata:
+        sample.metadata["rollout_id"] = rollout_id
```
zhuzilin: we can add the rollout_id (or `start_rollout_id` make be more accurate as the partial rollout sampling may contain segments of multiple rollouts) into metadata right before adding them back to the buffer.
That way, we don't need to add a `rollout_id` argument into `generate` and `generate_and_rm`.

slime/rollout/sglang_example.py
```diff
@@ -191,68 +309,80 @@ def submit_generate_tasks(samples):
...
+        for group_index, samples in data_group.items():
+            assert (
+                len(samples) == args.n_samples_per_prompt
+            ), f"Got {len(samples)} samples, expected {args.n_samples_per_prompt}"
+            # Only store samples that have reasonable partial content
+            cached_samples = []
+            for sample in samples:
+                # For half-completed or complete samples from incomplete groups, cache them
+                if sample.status != Sample.Status.PENDING:
+                    if (partial_rollout_filter is not None and not partial_rollout_filter(args, sample)):
```
zhuzilin: we can always add the samples back to the buffer and remove them using the `buffer_filter_path`,

guapisolo: > we can always add the samples back to the buffer and remove them using the `buffer_filter_path`,

Got it. When I write the code, I never consider the situation when train engine and rollout engine are fully async.
So the staleness info should be maintained in the buffer_filter?

slime/rollout/sglang_example.py
guapisolo: > maybe changing the `sample.status` into `sample.finish_reason` and do:
> 
> ```python
> sample.finish_reason = Sample.FinishReason[
>     output["meta_info"]["finish_reason"]["type"].upper()
> ]
> ```
> 
> would be cleaner.

But Pending samples haven't start rollout yet. Use `finish_reason` may be confusing, setting to `None` may cause bugs.

slime/rollout/sglang_example.py
guapisolo: > the reason for having a retry is that some time sglang will be inavailable for a moment. so we don't need to add a max_retries here. If the worker died, we should stop the training.

But in the original version, if SGLang engine has dead, the generate function will keep sending rollout request to SGLang and stuck in the loop..

zhuzilin: hmm... yeah, it makes sense. Then let's keep it this way and make the max_retries 60 so that it will last at least 60s.

slime/rollout/sglang_example.py
zhuzilin: make sense. let's keep the origin implementation.

slime/rollout/sglang_example.py
zhuzilin: > staleness info should be maintained in the buffer_filter

I think we can same the state of the sample into `Sample.metadata`, and pass the currently rollout id into buffer_filter.

PR #5, opened 2025-06-20
refactor: simplify timer.py

GeeeekExplorer:


zhuzilin:
wow, this is so much cleaner!

PR #15, opened 2025-06-28
update the doc

zlH518:
When I try to use this script to convert the model parameters on huggingface, slime-plugins cannot be found, so it is best to explicitly indicate the path where slime exists.


Chen-GX:
I suggest using `cd /your_path/slime && pip install -e .` and checking the startup path of your convert script to solve this problem, rather than using a hard-coded path(`/root/slime/`) in`PYTHONPATH`.

zhuzilin:
I agree with @Chen-GX . It would be better to `pip install -e .` in advance.

zlH518:
Thank you very much for the answer

PR #24, opened 2025-07-03
Add support for `reinforce++` and `reinforce++-baseline`

### LINKED ISSUE #16: [feat] Supporting more RL algorithms.
Thanks to your astonishing RL framework ❤️. I was wondering if you have any plan to support more RL algorithms, such as reinforce++.

Chen-GX:
## Summary
This PR introduces support for the `reinforce++` and `reinforce++-baseline` algorithms into the Slime framework. It aims to evaluate their performance in comparison to the existing `grpo` implementation.

Closes #16 

## Key Changes

- Implemented the core logic for `reinforce++` and its baseline variant in `ppot_utils.py`.
- Implemented the `distributed_masked_whiten` for normalization in `slime/utils/distributed_utils.py`.
- Added new experiment scripts: `scripts/run-qwen3-4B-rf.sh` and `scripts/run-qwen3-4B-rf-baseline.sh`.

## Experimental Results & Observations

I conducted experiments comparing the three algorithms, with results logged to wandb.

As shown in the charts below, `reinforce++-baseline` and `grpo` exhibit similar performance curves. However, the `reinforce++` implementation shows a distinct behavior (e.g., its reward curve appears more volatile/lower).

I have double-checked my implementation, and it seems to align with the algorithm's principles. I suspect the difference might stem from hyperparameter sensitivity or a subtle implementation detail.

I would appreciate it if you could pay special attention to the logic in `ppot_utils.py` and `slime/utils/distributed_utils.py`. I'm ready to make corrections if any mistakes are found.

BTW, this PR focuses on `reinforce++` and `reinforce++-baseline`. I'm happy to contribute implementations for other algorithms, such as RLOO, in a future PR if there is interest.

image
image
image
image

### DIFF
--- a/slime/rollout/sglang_example.py
+++ b/slime/rollout/sglang_example.py
@@ -74,21 +74,26 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
         response_token_ids = state.tokenizer(sample.response, add_special_tokens=False)["input_ids"]
         sampling_params["max_new_tokens"] -= len(response_token_ids)
 
-    if sampling_params["max_new_tokens"] > 0:
-        # Handle partial rollout samples: continue generation from existing response
-        input_text = sample.prompt + sample.response
+    assert (
+        sampling_params["max_new_tokens"] >= 0
+    ), f"max_new_tokens: {sampling_params['max_new_tokens']} should not be less than 0"
+    if sampling_params["max_new_tokens"] == 0:
+        return sample
 
-        payload = {
-            "text": input_text,
-            "sampling_params": sampling_params,
-        }
+    # Handle partial rollout samples: continue generation from existing response
+    input_text = sample.prompt + sample.response
 
-        output = await post(url, payload, use_http2=args.use_http2)
-        sample.response += output["text"]
+    payload = {
+        "text": input_text,
+        "sampling_params": sampling_params,
+    }
 
-        if output["meta_info"]["finish_reason"]["type"] == "abort":
-            sample.status = Sample.Status.ABORTED
-            return sample
+    output = await post(url, payload, use_http2=args.use_http2)
+    sample.response += output["text"]
+
+    if output["meta_info"]["finish_reason"]["type"] == "abort":
+        sample.status = Sample.Status.ABORTED
+        return sample
 
     prompt_tokens_ids = state.tokenizer(sample.prompt, add_special_tokens=False)["input_ids"]
     response_token_ids = state.tokenizer(sample.response, add_special_tokens=False)["input_ids"]
@@ -144,7 +149,7 @@ async def generate_and_rm_group(args, group: list[Sample], sampling_params: dict
         return group
 
     group = await asyncio.gather(
-        *[generate_and_rm(args, sample, sampling_params, evaluation=evaluation) for sample in group]
+        *[generate_and_rm(args, sample, sampling_params.copy(), evaluation=evaluation) for sample in group]
     )
 
     # for the rm that need the whole group, we will not do the rm here


slime/backends/megatron_utils/data.py
```diff
@@ -280,6 +280,15 @@ def process_rollout_data(rollout_id, args, data_buffer):
...
         data["rewards"] = rewards
 
+    elif args.advantage_estimator == "reinforce_plus_plus_baseline" and args.rewards_normalization:
+        # group baseline subtraction for RF++-baseline
+        rewards = torch.tensor([r for r in rewards], dtype=torch.float)
+        rewards = rewards.reshape(-1, args.n_samples_per_prompt)
+        mean = rewards.mean(dim=-1, keepdim=True)
+        rewards = rewards - mean
+        rewards = rewards.flatten().tolist()
+        data["rewards"] = rewards
```
zhuzilin: here we can merge the code with the abort normalization, e.g.

```python
if args.rewards_normalization:
    ...
    if args.advantage_estimator == "grpo" and args.grpo_std_normalization:
        std = rewards.std(dim=-1, keepdim=True)
        rewards = rewards / (std + 1e-6)
    ...
```

slime/backends/megatron_utils/loss.py
```diff
@@ -147,6 +149,27 @@ def compute_advantages_and_returns(args):
...
+
+    elif args.advantage_estimator == "reinforce_plus_plus_baseline":
+        advantages = get_reinforce_plus_plus_baseline_advantages(
+            rewards=rewards,
+            kl=kl,
+            loss_masks=loss_masks,
+            response_lengths=response_lengths,
+            kl_coef=args.kl_coef,
+        )
+        returns = advantages
```
zhuzilin: could you check if the result is the same with CP > 1. It may be a little tricky to implement the context parallel version for those 2 algos.

slime/utils/ppo_utils.py
```diff
@@ -119,3 +121,111 @@ def get_grpo_returns(
...
+    # Broadcast to get unwhitened advantages
+    unwhitened_advantages = [
+        torch.ones_like(kl_tensor) * reward_val - kl_coef * kl_tensor
+        for kl_tensor, reward_val in zip(kl, rewards)
+    ]
+    # Concatenate tensors for a global operation
+    if loss_masks is None:
+        loss_masks = [
+            torch.ones_like(adv) for adv in unwhitened_advantages
+        ]
```
zhuzilin: I've removed the case where `loss_masks is None`, we don't need this anymore.

slime/utils/ppo_utils.py
```diff
@@ -119,3 +121,111 @@ def get_grpo_returns(
...
+
+    all_advs = torch.cat(unwhitened_advantages)
+    all_masks = torch.cat(loss_masks)
+    
+    whitened_advs_flat = distributed_masked_whiten(all_advs, all_masks, shift_mean=True)
```
zhuzilin: Can we move the `distributed_masked_whiten` to the `NotImplementedError` in `if args.normalize_advantages:`?

slime/backends/megatron_utils/data.py
Chen-GX: Thanks for the great suggestion! I've refactored the reward normalization logic to merge the common parts for grpo and reinforce_plus_plus_baseline, eliminating the duplicated code. This should be much cleaner and easier to maintain now.

slime/utils/ppo_utils.py
Chen-GX: Thanks for your reminder. I've removed it.

slime/backends/megatron_utils/loss.py
Chen-GX: I've added the support for comtext parallel, and I'll test it before pr.

slime/utils/ppo_utils.py
Chen-GX: I've moved the distributed_masked_whiten logic to the if args.normalize_advantages: block in compute_advantages_and_returns. Accordingly, the helper functions now only return the unwhitened values. This makes the control flow much cleaner.

PR #26, opened 2025-07-03
Tiny remove two flush caches

fzyzcjy:
Hi, when glancing at code, I find there seems to be a bit of redundant flush cache:

For release_memory_occupation, I already called flush_cache in SGLang

image

For resume_memory_occupation, if I understand correctly, SGLang do not run any req during sleeping, so no cache need to be flushed

(Note: I do not test this PR locally since not running slime yet)

### DIFF
--- a/slime/backends/sglang_utils/sglang_engine.py
+++ b/slime/backends/sglang_utils/sglang_engine.py
@@ -94,6 +94,8 @@ class SglangEngine:
         self.llm.flush_cache()
 
     def sleep(self, level=1):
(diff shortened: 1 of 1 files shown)


zhuzilin:
The `flush_cache` here in the `sleep` function is mainly used as a probe to ensure there are no currently running requests on sglang; it's a relatively ad-hoc solution. The `flush_cache` within `wake_up` call can indeed be removed.

fzyzcjy:
I see, looks reasonable!

fzyzcjy:
updated

PR #41, opened 2025-07-08
refactor: major refactor for memory saving

zhuzilin:
This PR tries to diminish the memory occupancy of megatron:
1. use gloo for most gather_object and all_gather_object
~2. add an warmup to the model initialization~ The warmup doesn't solve the issue completely.
3. only initialize one model and make ref or old_actor as cpu state_dict

### DIFF
--- a/slime/backends/megatron_utils/model.py
+++ b/slime/backends/megatron_utils/model.py
@@ -337,7 +337,10 @@ def should_disable_forward_pre_hook(args):
     return not args.use_custom_fsdp and args.use_distributed_optimizer and args.overlap_param_gather
 
 
-def setup_before_train(args, model, optimizer):
+def train(rollout_id, model, optimizer, opt_param_scheduler, data_iterator, num_microbatches):
+    """Training function: run train_step desired number of times."""
+    args = get_args()
+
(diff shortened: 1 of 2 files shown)


slime/backends/megatron_utils/model.py
```diff
@@ -110,10 +109,99 @@ def setup_model_and_optimizer(
         skip_load_to_model_and_opt=False,
     )
 
+    warmup(args, model, optimizer)
+
     print(f"Loaded checkpoint at iteration {iteration}")
     return model, optimizer, opt_param_scheduler, iteration
 
 
+def warmup(args, model, optimizer):
```
fzyzcjy: curious whether we can put all training steps under mem_saver.region()

zhuzilin: yeah, I believe this is a better idea than `warmup`. I will remove the warmup here and try to put the whole training procedure under `mem_saver.region()`.

fzyzcjy: sure! if you find any errors just ping me (e.g. I expect some tensors may be used for cross-gpu communication and you need torch_memory_saver.disable() on it to avoid crashes).

PR #60, opened 2025-07-14
Fix/bug

zlH518:
I've observed that in both colocated and separated configurations, the num_gpus_per_actor for the actor model and the rollout engine consistently uses a 0.8 and 0.2 combination, respectively. This configuration is impacting inference throughput.

beside, i also add wandb offline mode in some private env


zhuzilin:
The 0.2 and 0.8 won't affect the actual GPU allocation, because ray only does logical allocation.

zlH518:
> The 0.2 and 0.8 won't affect the actual GPU allocation, because ray only does logical allocation.0.2 和 0.8 不会影响实际的 GPU 分配，因为 ray 只进行逻辑分配。

Thanks for your advice！

PR #65, opened 2025-07-15
Add loss mask when logging logp

yitianlian:
In previous logs of log-probabilities (logp) and reference logps, we used the mean over the entire output. However, we only care about the model’s generated output, so we need to apply a loss mask when logging these values.


slime/backends/megatron_utils/data.py
```diff
@@ -353,13 +355,56 @@ def log_rollout_data(rollout_id, args):
...
-                        # multiply by the cp_size.
-                        val = sum([cp_size * v.sum() / l for v, l in zip(val, response_lengths)]) / len(val)
+                        # Original logic for other tensor values
+                        if cp_size == 1:
+                            val = sum([v.mean() for v in val]) / len(val)
+                        else:
+                            # When cp_size > 1, the denominator should be the length of the response lengths. Also, to make
+                            # sure these values can be divided by `mpu.get_data_parallel_world_size(with_context_parallel=True)`
+                            # multiply by the cp_size.
+                            val = sum([cp_size * v.sum() / l for v, l in zip(val, response_lengths)]) / len(val)
```
zhuzilin: can we use the `get_sum_of_sample_mean` in the cp_util?

Issue #80, opened 2025-07-18
Brainstorms about code refactor


fzyzcjy:
Here is a simple list of brainstorms of code refactors to improve code clarity

- [ ] Refactor the Buffer
- [ ] Use hydra or something similar
- [ ] TODO: carefully read code

guapisolo:
Hi Tom, I am interested in the buffer refractor work and got some basic idea about that. Maybe we can discuss it later.

fzyzcjy:
@guapisolo sure, you can also find me on sglang slack btw

zlH518:
I also want to join this work, how can I contact you to get more details, such as development documents?

fzyzcjy:
You can find me at slack for details of the refactor thoughts; but maybe contact @zhuzilin since he is the owner of this repo for decisions like collaborations etc.

However, refactoring needs thorough understanding of the framework and programming methodologies. Therefore, if you are new to the field, I personally suggest starting with other small tasks may be a good choice.

zlH518:
> You can find me at slack for details of the refactor thoughts; but maybe contact [@zhuzilin](https://github.com/zhuzilin) since he is the owner of this repo for decisions like collaborations etc.关于重构的详细想法，你可以在 slack 上找到我；但也许应该联系他，因为他是这个仓库的所有者，对于合作等决策有决定权。
> 
> However, refactoring needs thorough understanding of the framework and programming methodologies. Therefore, if you are new to the field, I personally suggest starting with other small tasks may be a good choice.然而，重构需要对框架和编程方法有深入的理解。因此，如果你是新手，我个人建议从其他小任务开始可能是个不错的选择。

Thank you very much, I'm a complete novice：）

PR #84, opened 2025-07-21
refactor: remove global var LOCAL_STORAGE

zhuzilin:
relates to #80

### DIFF
--- a/slime/backends/utils/data.py
+++ b/slime/backends/utils/data.py
@@ -11,11 +11,11 @@ from slime.utils.timer import Timer
 class DataIterator:
     def __init__(
         self,
-        local_storage,
+        rollout_data,
         micro_batch_size: Optional[int] = None,
         micro_batch_indices: Optional[list[list[int]]] = None,
     ):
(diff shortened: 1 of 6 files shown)


slime/backends/utils/data.py
```diff
@@ -1,47 +1,21 @@
...
-
-
-def get_metadata(key: Optional[str] = None):
-    return LOCAL_METADATA.get(key, None)
-
-
 class DataIterator:
     def __init__(
         self,
+        local_storage,
```
fzyzcjy: nit: wondering whether renaming `local_storage` will be a tiny little bit better

zhuzilin: do you have any suggestion for the name?

fzyzcjy: hmm one name may be "batch" or "data_batch" or "batch_data", if this does not confuse with other existing names (I have not checked)

zhuzilin: renamed to `rollout_batch`.

PR #105, opened 2025-07-26
Add fp8 per block quant triton kernel

hebiao064:
E2E Speed up for Qwen3-30B-A3B-FP8

From 22 seconds to 20 seconds, we should see more gain for larger model like 671B and 235B 

#### Before: It took 397us for the tensor quantization.
Screenshot 2025-07-27 at 12 06 14 PM

#### After: It took 173us for the tensor quantization
Screenshot 2025-07-27 at 12 06 58 PM


hebiao064:
Providing a profiling code diff which can be converted into a PR later: https://gist.github.com/hebiao064/a5f4a7e96b63287508d1d69ebc0f2678

hebiao064:
closed since https://github.com/THUDM/slime/pull/147 was merged

PR #133, opened 2025-08-03
Naive token in token out implement for sglang example

yitianlian:


slime/rollout/sglang_example.py
```diff
@@ -38,6 +38,13 @@ def __init__(self, args):
             no_stop_trim=True,
             spaces_between_special_tokens=False,
         )
+
+        # Add token output mode configuration
+        if getattr(args, "use_token_output", False):
+            self.use_token_output = True
+        else:
+            self.use_token_output = False
+        print(self.sampling_params)
```
zhuzilin: please remove the print here or format the print with messages.

also, since we have store the `self.args`, it seems that using `self.args.use_token_output` directly would be cleaner.

slime/rollout/sglang_example.py
```diff
@@ -81,20 +89,48 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
         sample.status = Sample.Status.TRUNCATED
         return sample
 
-    # Handle partial rollout samples: continue generation from existing response
-    input_text = sample.prompt + sample.response
+    if state.use_token_output:
```
zhuzilin: we have `args` passed to this function, so `args.use_token_output` would be better.

slime/rollout/sglang_example.py
```diff
@@ -81,20 +89,48 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
...
+            "text": input_text,
+            "sampling_params": sampling_params,
+        }
+
+        output = await post(url, payload, use_http2=args.use_http2)
+        sample.response += output["text"]
+        prompt_tokens_ids = state.tokenizer(sample.prompt, add_special_tokens=False)["input_ids"]
+        response_token_ids = state.tokenizer(sample.response, add_special_tokens=False)["input_ids"]
+        sample.tokens = prompt_tokens_ids + response_token_ids
+        sample.response_length = len(response_token_ids)
```
zhuzilin: we can share most of the code with if and else, e.g.

```python
    payload = {
        "text": input_text,
         "return_logprob": args.use_token_output,
    }
    if args.use_token_output:
        payload["input_ids"] = ...
    else:
        payload["text"] = ...
    output = await post
    if args.use_token_output:
        respone_token_ids = ...
        response = ...
    else:
        response = ...
        respone_token_ids = ...
    ...
```

PR #154, opened 2025-08-10
Flatten tensor

hebiao064:


hebiao064:
closed due to too many conflicts, will create a new one

PR #166, opened 2025-08-13
Move entrypoints from outside package to inside

fzyzcjy:


fzyzcjy:
close after offline discussions

PR #193, opened 2025-08-15
Remove unused args, merge train.py and train_async.py

rysaya:
This PR include:
1. merge `train_async.py`, `train.py`, `--offload` and `--colocate` to one `--colocate`. Reason: I think no one would make the async train with `--colocate` and make the sync train without `--colocate`. So they could strictly merge. While for `--colocate` and `--offload`. When using colocate, one must enable offload, while there are only very few scenario one would use offload without colocate (What I could get is deepspeed/FSDP, but it's also a bad idea to use 'heavy offload + no colocate' compared with 'light offload + colocate'). While we focus on megatron, I think we could safely merge them
2. Remove `--rollout-global-dataset`, `--update-weights-interval`, `--keep-old-actor`. I think they are lagacy args and no one is using them now(Update weight interval could easily controled by Rollout-batch-size and Train-batch-size)
4. Move some args 'if' conditions outside, which are easier for the user to control.


rysaya:
Don't know why the merge include so many commit infos, I think we could edit the merge commit info when doing the merge :)

slime/ray/placement_group.py
```diff
@@ -118,5 +118,8 @@ def create_actor_group(args, pg, wandb_run_id):
     return actor_model
 
 
-def create_rollout_manager(args, pg, wandb_run_id):
-    return RolloutManager(args, pg, wandb_run_id=wandb_run_id)
+def create_rollout_manager(args, pg, actor_model, wandb_run_id):
+    rollout_manager = RolloutManager(args, pg, wandb_run_id=wandb_run_id, init_gen_engines=not args.debug_train_only)
+    if not args.debug_train_only and not args.debug_rollout_only:
+        ray.get(actor_model.async_init_weight_update_connections(rollout_manager))
```
rysaya, line 124: The reason I make `init_gen_engines` arg and put `ray.get(actor_model.async_init_weight_update_connections(rollout_manager))` from outside to here is for future offlineDataController usage (Eg: SFT, DPO, BT-RM), while they don't need to init the gen engine and make  weight update connections

rysaya:
closed after offline discussion

PR #216, opened 2025-08-25
[feat] support generate multi samples for one prompt

zhuzilin:


casper-hansen:
@zhuzilin can you provide a guide on multi-agent?

zhuzilin:
@casper-hansen This is something we are still exploring. And we will add an example when we have an ideal workflow :)

PR #217, opened 2025-08-25
[Feature] Abort requests before offload model weights

yitianlian:


jhinpan:
p.s. : In order to solve illegal memory issue like this one:
image

PR #232, opened 2025-08-26
add rollout metrics

zyzshishui:
Add rollout_time, tokens_per_second_per_gpu, total_tokens, partial_samples, off_policy_ratio
Note that currently we can only compute tokens_per_second_per_gpu dividing by total data collection time, which include the time for reward computation. This is not accurate especially for scenarios where the reward computation is costly. Waiting for a better solution ;-)

### DIFF
--- a/slime/backends/utils/data.py
+++ b/slime/backends/utils/data.py
@@ -145,7 +145,7 @@ def process_rollout_data(args, rollout_data_ref, dp_rank, dp_size):
             torch.tensor(
                 slice_log_prob_with_cp(log_prob, total_length, response_length),
                 device=torch.cuda.current_device(),
-                dtype=torch.bfloat16,  # TODO: hardcode to bf16 at the moment
+                dtype=torch.float32,
             )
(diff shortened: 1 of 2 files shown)


slime/utils/types.py
```diff
@@ -20,6 +20,7 @@ class Sample:
     reward: Optional[Union[float, dict[str, Any]]] = None
     loss_mask: Optional[list[int]] = None
     rollout_log_probs: Optional[list[float]] = None  # Log probabilities from rollout engine
+    completion_tokens: Optional[int] = None
```
zhuzilin: could you share the difference between `tokens` and `completion_tokens`?

zyzshishui: `tokens` stores the entire response of the prompt; `completion_tokens` represents the number of tokens generated in the **current** iteration, which is provided by sglang directly. I also checked relevant docs and discussions in sgl-project and it seems exactly the same with the length of generated tokens.

PR #236, opened 2025-08-29
[feature] add --tis-clip-low

zhuzilin:


yaof20:
Hi Zilin, thanks for adding new features to TIS!

Actually, we have tried such "clip low" strategy in our `DAPO-32B` experiments and we have the following observations (we didn't mention in the blog) to share with you here:

(1) Naively adding a clip_low value of `1/C` does not bring any gain and even hurts the downstream performance;

(2) Adding a clip_low vlaue of `1/C` when the advantage is positive can bring some marginal gain.

Based on our observations, we may recommend not to use the low_clip (currently setting it as 0 by default is fine) at least under our experiment setting. 

We haven't tested it universally due to the resource limitation. We are wondering if you have any positive experimental results on this so that we can learn more from you. We are open to and welcome further discussions!

Thanks again for actively maintaining the TIS feature!

PR #237, opened 2025-08-29
retool in one commit

maocheng23:
Create retool example in slime, 

For SFT and RL screenshots:

image
image
image

### DIFF
--- a/examples/retool/README.md
+++ b/examples/retool/README.md
@@ -34,8 +34,21 @@ hf download font-info/qwen3-4b-sft-SGLang-RL --local-dir /root/font-info/qwen3-4
 ```
 
 2. Create torch dict
+For SFT 
 ```bash
-source scripts/models/qwen3-4B-2507.sh
+```bash
+source scripts/models/qwen3-4B.sh
+PYTHONPATH=/root/Megatron-LM python tools/convert_hf_to_torch_dist.py \
+    ${MODEL_ARGS[@]} \
+    --hf-checkpoint /root/Qwen/Qwen3-4B-Instruct-2507 \
+    --rotary-base 5000000 \
+    --save /root/Qwen/Qwen3-4B-Instruct-2507_torch_dist
+```
+
+Or RL only
+```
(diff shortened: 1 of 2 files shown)


scripts/models/qwen3-4B-2507.sh
```diff
@@ -0,0 +1,17 @@
...
+   --num-query-groups 8
+   --use-rotary-position-embeddings
+   --disable-bias-linear
+   --normalization "RMSNorm"
+   --norm-epsilon 1e-6
+   --rotary-base 5000000
+   --vocab-size 151936
+   --kv-channels 128
+   --qk-layernorm
+)
```
zhuzilin: we don't really need this. you can still source the origin 4B config and set `--rotary-base 5000000`

examples/retool/run_qwen3_4B_RL_origindata.sh
```diff
@@ -0,0 +1,158 @@
...
+   ${ROLLOUT_ARGS[@]} \
+   ${OPTIMIZER_ARGS[@]} \
+   ${GRPO_ARGS[@]} \
+   ${DISTRIBUTED_ARGS[@]} \
+   ${WANDB_ARGS[@]} \
+   ${PERF_ARGS[@]} \
+   ${EVAL_ARGS[@]} \
+   ${SGLANG_ARGS[@]} \
+   ${MISC_ARGS[@]} \
+   ${CUSTOM_ARGS[@]}
```
zhuzilin: I think we only need to leave one script that match the comparing blog. having 3 of them is kind of confusing.

examples/retool/retool_sft_qwen3_4b.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
```
zhuzilin: please rename this file as `retool_qwen3_4b_sft.sh` and maybe the rl file as `retool_qwen3_4b_rl`?

PR #242, opened 2025-09-01
[Feature] Support token in token out for multi turn tasks

yitianlian:
## Motivation  
Using reinforcement learning (RL) to improve LLM performance on agentic tasks is becoming increasingly important. However, the current implementation uses a text-in–text-out interface for simplicity. This design introduces token retokenization mismatches, which can cause training instability.  

## Proposed Solution  
To address this, I introduce a **radix tree** component that transparently manages rollout trajectories. The radix tree can store each task’s different trajectories (e.g., GRPO), including:  
- Raw text  
- Rollout log-probabilities  
- Original output token lists  

This component is invisible to end users. When a rollout is finished, it guarantees retrieval of the correct token list without exposing internal details.  

## Implementation Details  
- Implemented a radix tree that:  
  - Accepts text and tokens, and stores them in the tree.  
  - Finds the longest prefix to retrieve previous token lists for token-in–token-out usage.  
  - Allows insertion of new text sequences.  
  - Performs automatic cleanup of stale nodes for efficient memory usage.  

## Next Steps / TODO  
- Investigate how to integrate the radix tree seamlessly with the current Slime implementation.  
- Validate stability improvements during RL training.  
- Benchmark performance and memory usage under large-scale rollouts.


slime/rollout/sglang_rollout.py
yitianlian, line 1: Can we add a new rollout file as an example of generating with the slime router? The current sglang_rollout file has too much content. (We can discuss it later.)

zhaochenyang20:
@ChangyiYang rebase with main please. And, I think the best way is too validate with Z.ai

PR #253, opened 2025-09-03
add shell script for qwen3-32B task

Gao016:
Fix bug:
In /scripts/models/qwen3-32B.sh, add a new line: --untie-embeddings-and-output-weights. Otherwise, raise AssertionError: tie_word_embeddings in hf config False is not equal to untie_embeddings_and_output_weights False, please check the config.

Add file:
Add /scripts/run-qwen3-32B.sh for qwen3-32B tasks. the raw_reward values of five times rollout is shown in the figure:
image]

### DIFF
--- a/scripts/models/qwen3-32B.sh
+++ b/scripts/models/qwen3-32B.sh
@@ -14,4 +14,5 @@ MODEL_ARGS=(
    --vocab-size 151936
    --kv-channels 128
    --qk-layernorm
+   --untie-embeddings-and-output-weights
 )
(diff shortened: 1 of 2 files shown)


Zhangmj0621:
@zhuzilin Same error occurs when use qwen3-14B. Need to add same line in slime/scripts/models/qwen3-14B.sh. I also worry if other model has same issue.

zhuzilin:
@Zhangmj0621 hmm... In fact, this is not really an issue, we can always add `--untie-embeddings-and-output-weights` in the scripts to overwrite this setting.

PR #277, opened 2025-09-05
support llama eagle3

zyzshishui:
#### 1. Context Length Constraint
`sglang-EAGLE-LLaMA3-Instruct-8B` only support a 2048 context length, which makes target model's context also limited to 2048 (need to solve on sglang side, or change another draft model). Consequently, the combined length of prompt + response cannot exceed 2048 tokens.
#### 2. Draft Model Training Not Supported
The current implementation does not support training of the draft model itself. The draft model is only used in inference.
#### 3. Performance Issues
Under the current configuration, speculative decoding does not provide performance benefits and even result in negative performance impact :(


zhuzilin:
close as it seems that we can use external eagle3 draft model at the moment.

PR #279, opened 2025-09-05
Fix Hf model to Mcore checkpoint conversion on AMD gpus

gramesh-amd:
Problem: On AMD GPUs, the `tools/convert_hf_to_torch_dist.py` script was getting stuck during conversion from HuggingFace checkpoint to Megatron-Core format, resulting in incomplete checkpoints:

```bash
PYTHONPATH=${MEGATRON_LM_PATH} python tools/convert_hf_to_torch_dist.py --swiglu    --num-layers 36    --hidden-size 2560    --ffn-hidden-size 9728    --num-attention-heads 32    --group-query-attention    --num-query-groups 8    --use-rotary-position-embeddings    --disable-bias-linear    --normalization "RMSNorm"    --norm-epsilon 1e-6    --rotary-base 1000000    --vocab-size 151936    --kv-channels 128    --qk-layernorm    --no-gradient-accumulation-fusion    --use-cpu-initialization    --hf-checkpoint /root/Qwen3-4B    --save /shared-aig/goramesh/model/Qwen3-4B_mcore
```

- Conversion process hangs after model loading
- Incomplete checkpoint saved with only partial files: 
```
__0_0.distcp  __0_1.distcp  common.pt
```

Temporary Solution:

- Use `--use-cpu-initialization` flag
- Move model to CPU before saving checkpoint

```python
bridge.load_weights(model, hf_model_path)
print(f"Model loaded: {hf_model_path}")

if args.use_cpu_initialization:
    model[0] = model[0].cpu()
```

- Results in complete checkpoint:
```
latest_checkpointed_iteration.txt  release/
__0_0.distcp  __0_1.distcp  common.pt  metadata.json
```


gramesh-amd:
@yushengsu-thu Can you please review this?

gramesh-amd:
https://github.com/THUDM/slime/blob/c64c26062a9c7c70c0772c3aef3fac62fc9b3df0/slime/backends/megatron_utils/model.py#L482-L495

I also think we will have to do the same cpu conversion here before save_checkpoint to save ckpts during training on AMD GPUs

I can make that change and test it after discussing hf to mcore changes first

yushengsu-thu:
@gramesh-amd , 

1. Could you please modify the `slime/slime/backends/megatron_utils/model.py` and to make ckpt saving work well in this PR

2. Could you also modify the corresponding scripts/doc here: [run-qwen3-4B-amd.sh](https://github.com/THUDM/slime/blob/main/scripts/run-qwen3-4B-amd.sh), [amd_tutorial.md](https://github.com/THUDM/slime/blob/main/docs/en/platform_support/amd_tutorial.md)

3. After you re-solving checkpoint converting, checkpoint loading, checkpoint saving issues, could you also try on Qwen3-8B model (becasue Qwen3-4B has no checkpoint loading issue. Thus, we might miss this test)

cc. @zhuzilin

Issue #50, opened 2025-07-10
[feature] Auto-Configuring Model Arguments (MODEL_ARGS) from config.json


Chen-GX:
Currently, setting up a training run for a specific model requires manually specifying a large number of `MODEL_ARGS`. This process can be error-prone, especially for users who may not be familiar with the subtle architectural details of every model.

For example, the Qwen3-8B model requires the `--untie-embeddings-and-output-weights` flag to be set. If a user omits this flag, the training process completes without any errors, but the resulting model produces garbled/unusable output during inference. This "silent failure" can lead to a significant waste of time and computational resources.

Maybe we can develop a script, which could accept a path to the model's  config.json file, and then parse it to populate the arguments. For example, it could automatically determine:

```bash
--num-layers (from num_hidden_layers)
--hidden-size (from hidden_size)
--num-attention-heads (from num_attention_heads)
--ffn-hidden-size (from intermediate_size)
--seq-length (from max_position_embeddings)
--untie-embeddings-and-output-weights (based on the value of tie_word_embeddings in the config, which is false for Qwen3-8B)

and other relevant architectural parameters.
```
Users could still override these automatically inferred values with explicit command-line arguments if they need to experiment with different configurations.

zhuzilin:
close as we will gradually migrate to megatron-bridge for this.

PR #312, opened 2025-09-09
feat: auto configure megatron from hf config.

SanftMonster:
I like the way of initializing megatron in #50. Here's an implementation of it.

Currently only `Qwen2`, `Qwen3`, `Qwen3Moe` and `GLM4Moe` are added as predefined model types but it's easy to add more in the future. Could you please take a review of the design? @zhuzilin


zhuzilin:
Thank you so much for this! I have been hoping for this kind of support for a long time. There are a few nits to change but I believe we can merge the code first.

Issue #293, opened 2025-09-06
[FSDP] Implement data packing to reduce padding overhead


zhuzilin:
The current FSDP data pipeline groups samples into fixed batches with extensive padding, which leads to inefficient memory usage and slower training. To improve efficiency, we should implement data packing so that variable-length sequences can be packed together more tightly, reducing wasted computation from padding.

We would also need to support dynamic micro batch size adjustment at runtime to further improve efficiency when sequence length distribution varies. (`--use-dynamic-batch-size`).

cook-cmd:
maybe i can do it

jhinpan:
I'm working on this.

cook-cmd:
> I'm working on this.

oh ! maybe we can do it together?

cook-cmd:
oh ! may be we can do it  together? @jhinpan

jhinpan:
> oh ! may be we can do it together? [@jhinpan](https://github.com/jhinpan)

If you want, you are welcome to check if there is anything wrong in this PR #321

PR #321, opened 2025-09-10
[FSDP] Data Packing Implementation in FSDP backend

jhinpan:
In order to solve #293.

The 4 files implement an efficient data packing solution for FSDP that addresses padding waste:

  Key Components:

  1. `arguments.py` - Added configuration flags:
    - --use-data-packing: Enable packing
    - --use-dynamic-batch-size: Dynamic micro batch adjustment
    - --max-tokens-per-gpu: Token budget control
    - --pack-efficiency-threshold: Minimum 80% efficiency target
  2. `data_packing.py` - Core packing logic:
    - `pack_sequences()`: Groups variable-length sequences using cumulative lengths (cu_seqlens) for varlen attention
    - `compute_optimal_batch_size()`: Binary search to find optimal batch size based on sequence distribution
    - Three packing strategies: static binning, dynamic token-budget, balanced partitioning
  3. `actor.py` - Integration points:
    - Lines 175-225: Replaced naive padding with efficient packing in `pad_and_move_to_device()`
    - Lines 424-461: Added `gather_log_probs_packed()` for packed sequence handling
    - Lines 209-224: Dynamic batch size adjustment based on efficiency history
  4. `test_data_packing.py` - Comprehensive test coverage

  How It Works:

  Without packing (old way):
  Batch: [seq1_____] [seq2___] [seq3______]  # ~50% wasted on padding

  With packing (new way):
  Packed: [seq1|seq2|seq3]  # Dense tensor with cu_seqlens=[0,4,7,10]

  Dynamic batch size: When enabled with `--use-dynamic-batch-size`, the system:
  - Monitors packing efficiency over recent batches
  - Adjusts micro batch size to maintain ~85% efficiency
  - Reduces batch size if sequences vary widely in length
  - Increases batch size when sequences are similar

  This reduces memory usage and computation waste from padding, enabling more efficient training especially with highly variable sequence lengths.


jhinpan:
Tests passed on my end.
Screenshot 2025-09-09 at 10 44 22 PM

slime/backends/fsdp_utils/data_packing.py
```diff
@@ -0,0 +1,436 @@
+"""
+Data packing utilities for FSDP backend to reduce padding overhead.
+Supports packing variable-length sequences and dynamic micro batch size adjustment.
+Integrates with slime's sequence length balancing utilities.
+"""
```
zhuzilin: this is unecessarily long.... and we don't need the pack efficiency threshold, please don't do pre-mature optimization

slime/backends/fsdp_utils/actor.py
```diff
@@ -130,37 +151,90 @@ def compute_log_prob(
...
-                        rollout_data["rewards"][i : i + self.args.micro_batch_size],
-                        dtype=torch.float,
-                        device=torch.cuda.current_device(),
-                    ),
-                    "raw_reward": rollout_data["raw_reward"][i : i + self.args.micro_batch_size],
-                }
+        rewards = rollout_data.get("rewards", None)
+        raw_rewards = rollout_data.get("raw_reward", None)
+        
+        if self.use_data_packing:
```
zhuzilin: we can default to use_data_packing and remove the origin branch.

slime/backends/fsdp_utils/data_packing.py
jhinpan: Already removed and trimmed under 200 lines.

slime/backends/fsdp_utils/actor.py
```diff
@@ -130,37 +151,90 @@ def compute_log_prob(
         with timer(f"{store_prefix}log_probs") and torch.no_grad():
             for batch in padded_batches:
                 with torch.autocast(device_type="cuda", dtype=torch.bfloat16):
-                    logits = self.model(input_ids=batch["tokens"]).logits
-                batch[f"{store_prefix}log_probs"] = gather_log_probs(logits, batch["tokens"])
+                    if "cu_seqlens" in batch:
+                        # Handle packed sequences
+                        logits = self.model(input_ids=batch["tokens"]).logits
```
zhuzilin: we need to pass cu_seqlens to model to make sure the attn is computed correctly.

slime/backends/fsdp_utils/arguments.py
```diff
@@ -27,6 +27,22 @@ class FSDPArgs:
...
+    # Data packing configuration
+    use_data_packing: bool = False  # Enable efficient data packing
+    max_seq_len: int = 8192  # Maximum sequence length
+    
+    # Dynamic batch size adjustment
+    use_dynamic_batch_size: bool = False  # Enable dynamic micro batch size
+    max_tokens_per_gpu: Optional[int] = None  # Maximum tokens per GPU
+    min_micro_batch_size: int = 1  # Minimum micro batch size
+    max_micro_batch_size: int = 64  # Maximum micro batch size  
+    target_efficiency: float = 0.85  # Target packing efficiency
```
zhuzilin: please delete the unused params

slime/backends/fsdp_utils/data_packing.py
```diff
@@ -0,0 +1,197 @@
...
+from slime.utils.seqlen_balancing import get_seqlen_balanced_partitions
+
+def pack_sequences(
+    tokens: List[List[int]],
+    loss_masks: List[List[int]],
+    rewards: Optional[List[float]] = None,
+    raw_rewards: Optional[List[Any]] = None,
+    max_seq_len: Optional[int] = None,
+    max_tokens_per_gpu: Optional[int] = None,
+    num_packs: Optional[int] = None,
```
zhuzilin, line 20: we can calculate the num pack within the function. and I believe we only need to return `indices: List[List[int]]` for this function so that the don't need to pass rewards and raw_rewards in the function.

slime/backends/fsdp_utils/arguments.py
jhinpan: Hmm I think these params have some use? Or do u think that we should just remove target_efficiency related setting in the data_packing?

For now:
- max_seq_len and max_tokens_per_gpu are always used in packing_config
- use_dynamic_batch_size is used to control dynamic batching
- min_micro_batch_size, max_micro_batch_size, and target_efficiency are only used when use_dynamic_batch_size=True
- micro_batch_size is used as the base batch size

I will remove current unused use_ring_attention configuration anyway.

slime/backends/fsdp_utils/data_packing.py
jhinpan: Good point! Refactored.

slime/backends/fsdp_utils/actor.py
```diff
@@ -126,40 +131,36 @@ def compute_log_prob(
...
+                        logits, batch["tokens"], batch["cu_seqlens"]
+                    )
 
     def pad_and_move_to_device(self, rollout_data):
+        # Pack sequences efficiently
         tokens = rollout_data["tokens"]
         loss_masks = rollout_data["loss_masks"]
 
         padded_batches = []
         for i in range(0, len(tokens), self.args.micro_batch_size):
```
zhuzilin: I think that when using `max_tokens_per_gpu`, we should no long split data with `args.micro_batch_size`, but with `args.global_batch_size`, e.g.

```python
for i in range(0, len(tokens), self.args.global_batch_size):
    batch_tokens = tokens[i : i + self.args.global_batch_size]
    batch_loss_masks = loss_masks[i : i + self.args.global_batch_size]
    if self.use_dynamic_batch_size:
        mbs = split_gbs_into_mbs(batch_tokens, batch_loss_masks, self.args.max_tokens)
    else:
        mbs = [(batch_tokens[j:j + self.args.micro_batch_size] for j in range(0, len(batch_tokens), self.args.micro_batch_size))]
    pack_mbs = [pack_sequence(m) for m in mbs]
```

UbeCc:
@codex review

UbeCc:
@codex review

slime/backends/fsdp_utils/actor.py
```diff
@@ -165,60 +176,73 @@ def compute_log_prob(
...
+            for i in range(0, len(tokens), local_batch_size):
+                mbs_size_list.append(
+                    get_minimum_num_micro_batch_size(
+                        [len(t) for t in rollout_data["tokens"][i : i + local_batch_size]],
+                        self.args.max_tokens_per_gpu,
+                    )
+                )
+            num_microbatches = torch.tensor(mbs_size_list, dtype=torch.int, device=torch.cuda.current_device())
+            dist.all_reduce(num_microbatches, op=dist.ReduceOp.MAX)
+            num_microbatches = num_microbatches.tolist()
```
zhuzilin, line 201: I think it would be better if the if-else here can both return `num_microbatches`, we no longer need to rewrite the `packed_batches.extend(...)`.

```python
# we don't need to 
if self.args.use_dynamic_batch_size:
    ...
    num_microbatches = num_microbatches.tolist()
else:
    num_microbatches = [self.args.micro_batch_size] * (...)
grad_accum = list(accumulate(num_microbatches))

start = 0
for mbs in num_microbatches:
    end = start + mbs
    packed_batches.extend(pack_sequences(...))
    start = end
```

slime/backends/fsdp_utils/actor.py
```diff
@@ -165,60 +176,73 @@ def compute_log_prob(
...
+
+        assert (
+            self.args.global_batch_size % dp_size == 0
+        ), f"global_batch_size {self.args.global_batch_size} is not divisible by dp_world_size {dp_size}"
+        # Use global_batch_size for splitting when max_tokens_per_gpu is enabled
+        if (
+            hasattr(self.args, "max_tokens_per_gpu")
+            and self.args.max_tokens_per_gpu is not None
+            and self.args.use_dynamic_batch_size
+        ):
```
zhuzilin: we can assume the args has `args.use_dynamic_batch_size` and if `args.use_dynamic_batch_size` is set the `args.max_tokens_per_gpu is not None`.

slime/backends/fsdp_utils/actor.py
```diff
@@ -276,25 +304,39 @@ def train(self, rollout_id, rollout_data_ref):  # type: ignore[override]
...
+
+            old_log_probs = torch.cat([batch["log_probs"] for batch in unpacked_batches], dim=0)
+            log_probs = torch.cat([batch["cur_log_probs"] for batch in unpacked_batches], dim=0)
+            advantages = torch.cat([batch["advantages"] for batch in unpacked_batches], dim=0)
+            loss_masks = [batch["loss_masks"].to(device=log_probs.device) for batch in unpacked_batches]
+            response_lengths = [batch["response_lengths"] for batch in unpacked_batches]
+            ref_log_probs = (
+                torch.cat([batch["ref_log_probs"] for batch in unpacked_batches], dim=0)
+                if self.args.use_kl_loss
+                else None
```
zhuzilin: we can move this to the `if self.args.use_kl_loss` below.

slime/backends/fsdp_utils/actor.py
```diff
@@ -304,10 +346,10 @@ def train(self, rollout_id, rollout_data_ref):  # type: ignore[override]
             if self.args.use_kl_loss:
                 kl = compute_approx_kl(
                     log_probs,
-                    batch["ref_log_probs"],
+                    ref_log_probs.to(device=log_probs.device),
```
zhuzilin: this should always on current device?

slime/backends/fsdp_utils/actor.py
```diff
@@ -323,34 +365,27 @@ def train(self, rollout_id, rollout_data_ref):  # type: ignore[override]
             if self.args.use_kl_loss:
                 reported["kl_loss"] = kl_loss.detach()
 
-            loss = loss / grad_accum
+            # Scale loss for gradient accumulation
+            loss = loss * dist.get_world_size() / self.args.global_batch_size
             loss.backward()
-
+            clear_memory()
```
zhuzilin, line 352: hmm.. this could make the training really slow could you check if removing the clear_memory here will trigger oom?

slime/backends/fsdp_utils/arguments.py
```diff
@@ -26,6 +26,10 @@ class FSDPArgs:
     fsdp_sync_module_states: bool = True
     fsdp_forward_prefetch: bool = True
     fsdp_backward_prefetch: bool = True
+    attn_implementation: str = "flash_attention_2"
```
zhuzilin, line 29: let's just hardcode to `flash_attention_2` since this is the only attn impl that supports data packing.

slime/backends/fsdp_utils/arguments.py
```diff
@@ -26,6 +26,10 @@ class FSDPArgs:
...
+
+    # max_tokens_per_gpu is defined in main arguments (slime/utils/arguments.py)
+    micro_batch_size: int = 1  # Micro batch size
```
zhuzilin: I think that this arg should already been registered in `slime/utils/arguments.py`?

Issue #302, opened 2025-09-06
[FSDP] Verify FSDP backend availability via uv install / pip install


zhuzilin:
With the new FSDP backend integrated into slime, we need to confirm that users can directly install slime (including FSDP support) via `uv install` or `pip install` without additional manual steps. This ensures smooth adoption and reduces friction for users trying out the FSDP backend.

Williamren97:
Hi, Zilin, I'd like to take on this issue.

PR #325, opened 2025-09-11
[FSDP] Verify FSDP backend availability via uv install / pip install

Zhuohao-Li:
solve #302, add verification for fsdp backend in `fsdp_utils/__init__.py` similar to megatron style and verify by running `slime/test_fsdp.py`(temp)


slime/test_fsdp.py
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
zhuzilin: can we move this to `tests` folder and rename it to *.sh?

zhuzilin:
Thank you for this PR! Could you share how to build the environment with uv?

slime/test_fsdp.py
Zhuohao-Li: sorry for the typo, already fixed~

Issue #299, opened 2025-09-06
[FSDP] Add basic VLM data pipeline support


zhuzilin:
slime currently only supports text-only data pipelines. To enable vision-language model (VLM) training, we need a basic multimodal data flow that can handle paired image–text inputs.

PR #335, opened 2025-09-15
 Add basic VLM data pipeline

ppraneth:
This is my first PR to the project. 
It closes [FSDP] Add basic VLM data pipeline support #299 . Please let me know if I should improve anything.
I’m happy to make changes. Thanks for reviewing!


zhuzilin:
Thank you for the PR! I think the main issue is that we should not load the image from Buffer and send the pixel value through ray, that will involve too much communication. An ideal way would be sending the image file to the training part and let each GPU recreate the prompt.

ppraneth:
> Thank you for the PR! I think the main issue is that we should not load the image from Buffer and send the pixel value through ray, that will involve too much communication. An ideal way would be sending the image file to the training part and let each GPU recreate the prompt.

Thanks for pointing that out! I’ve updated the code
Could you please check it again and let me know if any further adjustments are needed?

slime/backends/megatron_utils/actor.py
```diff
@@ -182,6 +184,19 @@ def _get_rollout_data(self, rollout_data_ref):
...
+                if path:
+                    image=Image.open(path).convert("RGB")
+                    inputs=self.vlm_processor(images=image, return_tensors="pt")
+                    pixel_values=inputs.pixel_values.to(
+                        device=torch.cuda.current_device(), dtype=torch.bfloat16
+                    )
+                    pixel_values_list.append(pixel_values)
+                else:
+                    pixel_values_list.append(None)
+            rollout_data["pixel_values"] = pixel_values_list
```
zhuzilin: Could you move the changes in `backends/megatron_utils` into `backends/fsdp_utils`? Megatron does not natively support passing pixel_values into the model so we may need to support vlm through the still developing fsdp backend.

Also, I wonder if we have a more general way to pass different kinds of multimodel data, e.g. videos, audios.

ppraneth: I’ve moved the changes from backends/megatron_utils into backends/fsdp_utils as suggested. Thanks for pointing that out! Could you please review again and let me know if any further refinements are needed?

zhuzilin:
It looks fine now :) Let me run the CI to see if there are bugs with the origin text only training, for example I'm not sure if we can use `{"text": data[prompt_key], "type": "text"}` with llm.

ppraneth:
> It looks fine now :) Let me run the CI to see if there are bugs with the origin text only training, for example I'm not sure if we can use `{"text": data[prompt_key], "type": "text"}` with llm.

Thanks for the review and for running the CI! You were right. The CI failed with a Jinja template error, confirming that the new data structure for prompts was not backward compatible with the text-only training.
I've just pushed a new commit that fixes this.

The changes made in  slime/utils/data.py now conditionally handle the prompt formatting. If multimodal_keys isn't present, it leaves the prompt data in its original format for the text-only pipeline. The new list-based structure is only created when multimodal data is actually being used.

This should resolve the CI error. Could you please take another look and re-run the CI when you get a chance? Thanks again for catching that!

ppraneth:
@zhuzilin Why is the pre-commit failing?

zhuzilin:
@ppraneth oh, you need to do `pre-commit install` before the commit so that there will be automatic linting for the committed code. anyway, I think we can merge the code and I can help fix the linting.

tellyoung:
May I ask if there has a test shell script? Which models are supported by this multimodal system?

PR #340, opened 2025-09-16
add dockerfile and patch for b200

maocheng23:
As title, create b200 dockerfile and patches for sglang v0.5.2rc2-cu128-b200


zhyncs:
Hi all @zhuzilin @maocheng23 
After v0.5.2, we can use `lmsysorg/sglang:v0.5.2` for hopper and blackwell. And latest version is `v0.5.3rc0`. Thanks!

maocheng23:
Great to now that! Later we can use a single dockerfile for both hopper and blackwell!

Issue #296, opened 2025-09-06
[FSDP] Add reference model support for correct KL loss computation


zhuzilin:
Currently, the FSDP backend does not support a reference model, which makes it impossible to correctly compute KL loss (e.g., in RLHF or DPO training). Adding support for a reference model will enable correct and efficient KL loss calculation under FSDP.

To ensure reliability, the implementation should also be numerically aligned with the Megatron backend. (Similar to #290).

UbeCc:
maybe i can do it :)

zhuzilin:
fixed in #344

PR #344, opened 2025-09-17
[FSDP] Add reference model support for correct KL loss computation #296

UbeCc:
Here I have implemented the reference model support as mentioned in #296. Also, I add `wandb_secondary_init` as megatron backend for wandb logging (Otherwise it will through init error)

In the implementation, slime will load reference model using `AutoModelForCausalLM` once and store the params into cpu. When calculating KL, slime will reload the params and forward normally.


UbeCc:
I will make precision comparison between fsdp and megatron backend :)

UbeCc:
image

FSDP KL loss matches Megatron now ~ (plz note FSDP should make temperature shifting as Megatron does. Because the trajectory generated by the inference engine uses temperature while the `forward` operation by training engine does not )

UbeCc:
megatron script: https://pastebin.ubuntu.com/p/vsc8DP2v85/
fsdp script: https://pastebin.ubuntu.com/p/m4NtXn26zx/

PR #346, opened 2025-09-17
[FSDP] fix: fix fsdp llh calc div temperature to match megatron

jeffrey-dot-li:
Megatron LLH calculation divides logits by temperature, this fixes fsdp to match

https://github.com/THUDM/slime/blob/4badb749b213408bdf24d799d6c781b2d3c5c31f/slime/backends/megatron_utils/loss.py#L33


UbeCc:
Well it is implemented in https://github.com/THUDM/slime/pull/344. Both implementation is ok!

jeffrey-dot-li:
o nice thanks ill close then

PR #351, opened 2025-09-18
Fix ray get nested refs

jeffrey-dot-li:
Currently, `ray.get` calls for offload and onload (e.g.) [ray.get(rollout_manager.offload.remote())](https://github.com/THUDM/slime/blob/8f331dc6c2f01173e9f2fb17f709d1277642eb28/train.py#L69) do not wait for the nested ray calls for example ```
def offload(self):
        return [engine.release_memory_occupation.remote() for engine in self.rollout_engines]
 ```
 
 This means that the next lines of code (e.g. `actor_model.async_train`) will start possibly before the offload has completed. This causes non-deterministic / flakey OOMs as sometimes the actor_model will attempt to start allocating training memory before SGLang has finished offloading.


zhuzilin:
hmm... I'm afraid the origin code is correct. I wonder if you met the unexpected behavior.
image

jeffrey-dot-li:
> hmm... I'm afraid the origin code is correct. I wonder if you met the unexpected behavior. image

Can you try 

```
if args.rollout_global_dataset:
        ray_result = ray.get(rollout_manager.load.remote(args.start_rollout_id - 1))
        print(f"{ray_result=}")
```

if this is gathering the objects properly, it should be a list of Nones, but instead I get `ObjectRef`

zhuzilin:
Sorry for the late replay. Yeah you are correct, this was an issue in slime. Thank you so much for the PR. Close as it is seemed to be fixed.

PR #353, opened 2025-09-19
Added back --rollout-num-gpus-per-node for backward compatibility

sam571128:
Added back the flag since this broke my old training scripts. Feel like it makes sense to keep it as an alias as it is pretty clear that it indicates the num of gpus for rollout.


zhuzilin:
Really sorry the update broke your script... However we renamed the param to `--num-gpus-per-node` because it actually means how many GPUs are there on one node, for example, 8 gpu for h100 and 4 gpu for gb200 nvl72.

sam571128:
makes sense, thanks a lot

PR #355, opened 2025-09-19
feature: Add total progress bar

### LINKED ISSUE #311: Add Total Rollout Progress Bar for Training Monitoring
Currently, slime only displays progress bars for individual rollout generation, but lacks visibility into overall training progress across all rollouts.

This makes it difficult to estimate total experiment completion time for long-running training jobs.

ppraneth:
Closes #311


zhuzilin:
the tqdm is not suitable here as there will be lots of logs between each iteration. and I think that in the current rl training, we will rarely complete all the num rollout so that the progress bar here won't give much help.

PR #401, opened 2025-09-29
Include type annotation in update_weight_utils.py

lancerts:


lancerts:
@zhuzilin merge conflict resolved, ty

PR #404, opened 2025-09-29
add prompt filter with the prompt token length

none0663:
add prompt filter with the prompt token length for over length prompt


zhuzilin:
we should not use multi-thread for cpu intensive operation in python due to GIL.

PR #406, opened 2025-09-29
[FSDP] [feature]Support save/load checkpointing

### LINKED ISSUE #402: [FSDP] Support save/load checkpointing
For now, FSDP misses the saving and loading model like megatron which should be supported

For example:
Megatron:

```python
def save_model(self, iteration):
    if self.args.debug_rollout_only:
        return

    save(iteration, self.model, self.optimizer, self.opt_param_scheduler)
```

FSDP:
```python
def save_model(self, iteration):
    if self.args.debug_rollout_only:
        return

    raise NotImplementedError()
```

Better to save the model to be huggingface format.

ppraneth:
Hi everyone,

This pr add load/save feature in FSDP and also closes #402 

Here are the main changes I've made:

* **Checkpointing Logic**: I added all the logic for saving and loading into a new file, `slime/backends/fsdp_utils/checkpoint.py`, to keep things organized. This handles saving the model, tokenizer, and optimizer state.

* **Actor Integration**: I updated the `FSDPTrainRayActor` in `slime/backends/fsdp_utils/actor.py` to use the new functions. It now correctly loads the state from a checkpoint at the start of a run and saves progress when `save_model` is called.

* **New Arguments**: To control this, I added a few command-line arguments to `slime/backends/fsdp_utils/arguments.py`: `--save`, `--load`, `--save-safe-serialization` and a safety flag `--overwrite-checkpoints` so you don't accidentally overwrite your work.

* **Bug Fixes**: While building this, I also fixed a couple of bugs in the `init` process to prevent the model from being loaded twice and to make sure the `global_step` counter is restored correctly.

Please check it out and let me know if any changes are needed. I'm happy to make them! Thanks for taking a look.

### DIFF
--- a/slime/backends/fsdp_utils/actor.py
+++ b/slime/backends/fsdp_utils/actor.py
@@ -4,12 +4,12 @@ from itertools import accumulate
 import ray
 import torch
 import torch.distributed as dist
+import wandb
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.distributed.fsdp import ShardingStrategy, StateDictType
 from torch_memory_saver import torch_memory_saver
 from transformers import AutoConfig, AutoModelForCausalLM, AutoProcessor, AutoTokenizer
 
-import wandb
 from slime.ray.train_actor import TrainRayActor
 from slime.utils.data import get_minimum_num_micro_batch_size, process_rollout_data
 from slime.utils.distributed_utils import get_gloo_group
@@ -18,6 +18,7 @@ from slime.utils.ppo_utils import compute_approx_kl, compute_policy_loss
 from slime.utils.timer import Timer, timer
 from slime.utils.wandb_utils import init_wandb_secondary
 
+from .checkpoint import load_checkpoint, save_checkpoint
 from .data_packing import pack_sequences, unpack_sequences
 from .update_weight_utils import UpdateWeightFromDistributed, UpdateWeightFromTensor
 
@@ -44,10 +45,14 @@ class FSDPTrainRayActor(TrainRayActor):
         self.args = args
         torch.manual_seed(args.seed)
 
+        hf_checkpoint = args.load if args.load else args.hf_checkpoint
+        loaded_rollout_id = -1
+        self.global_step = 0
+
         for i in range(dist.get_world_size()):
             if i == dist.get_rank():
-                self.hf_config = AutoConfig.from_pretrained(self.args.hf_checkpoint, trust_remote_code=True)
-                self.tokenizer = AutoTokenizer.from_pretrained(self.args.hf_checkpoint, trust_remote_code=True)
+                self.hf_config = AutoConfig.from_pretrained(hf_checkpoint, trust_remote_code=True)
+                self.tokenizer = AutoTokenizer.from_pretrained(hf_checkpoint, trust_remote_code=True)
             dist.barrier(group=get_gloo_group())
 
         if self.args.multimodal_keys:
@@ -56,7 +61,7 @@ class FSDPTrainRayActor(TrainRayActor):
         # Load model
         with torch.autocast(device_type=f"cuda:{torch.cuda.current_device()}"):
             model = AutoModelForCausalLM.from_pretrained(
-                self.args.hf_checkpoint,
+                hf_checkpoint,
                 trust_remote_code=True,
                 attn_implementation=self.args.attn_implementation,
             )
@@ -89,7 +94,8 @@ class FSDPTrainRayActor(TrainRayActor):
             weight_decay=args.weight_decay,
         )
 
-        # TODO: load
+        if args.load:
+            loaded_rollout_id, self.global_step = load_checkpoint(args, self.model, self.optimizer)
 
         self.weights = {"actor": {}}
 
@@ -113,9 +119,8 @@ class FSDPTrainRayActor(TrainRayActor):
             self.sleep(("model"))
 
         Timer().start("train_wait")
-        self.global_step = 0
         self.micro_step = 0
-        return 0
+        return loaded_rollout_id + 1
 
     def sleep(self, tags):
         if not getattr(self.args, "offload", False):
@@ -133,7 +138,14 @@ class FSDPTrainRayActor(TrainRayActor):
         if self.args.debug_rollout_only:
             return
 
-        raise NotImplementedError()
+        save_checkpoint(
+            self.args,
+            iteration,
+            self.model,
+            self.optimizer,
+            self.tokenizer,
+            self.global_step,
+        )
 
     def compute_log_prob(
         self,
--- a/slime/backends/fsdp_utils/checkpoint.py
+++ b/slime/backends/fsdp_utils/checkpoint.py
@@ -1,0 +1,75 @@
+import os
+import torch
+import torch.distributed as dist
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp import StateDictType
+
+
+def save_checkpoint(args, iteration, model, optimizer, tokenizer, global_step):
+    """Save a checkpoint."""
+    if dist.get_rank() == 0:
+        checkpoint_dir = os.path.join(args.save, str(iteration))
+
+        if os.path.exists(checkpoint_dir):
+            if not args.overwrite_checkpoints:
+                print(
+                    f"WARNING: Checkpoint directory {checkpoint_dir} already exists and --overwrite-checkpoints is not set. "
+                    "Skipping saving."
+                )
+                return
+            else:
+                print(f"WARNING: Overwriting existing checkpoint directory {checkpoint_dir}.")
+
+        print(f"Saving checkpoint at iteration {iteration} to {checkpoint_dir}")
+        os.makedirs(checkpoint_dir, exist_ok=True)
+
+        with FSDP.state_dict_type(model, StateDictType.FULL_STATE_DICT):
+            state_dict = model.state_dict()
+
+        model.save_pretrained(
+            checkpoint_dir,
+            state_dict=state_dict,
+            safe_serialization=args.save_safe_serialization,
+        )
+        tokenizer.save_pretrained(checkpoint_dir)
+
+        torch.save(optimizer.state_dict(), os.path.join(checkpoint_dir, "optimizer.pt"))
+
+        training_state = {
+            "iteration": iteration,
+            "global_step": global_step,
+        }
+        torch.save(training_state, os.path.join(checkpoint_dir, "training_state.pt"))
+        print(f"Checkpoint saved to {checkpoint_dir}")
+
+    dist.barrier()
+
+
+def load_checkpoint(args, model, optimizer):
+    """Load a checkpoint into the provided model and optimizer."""
+    loaded_rollout_id = -1
+    global_step = 0
+    checkpoint_path = args.load
+    if not checkpoint_path:
+        return loaded_rollout_id, global_step
+
+    print(f"Loading checkpoint from {checkpoint_path}")
+
+    optimizer_path = os.path.join(checkpoint_path, "optimizer.pt")
+    if os.path.exists(optimizer_path):
+        optimizer_state = torch.load(optimizer_path, map_location="cpu")
+        optimizer.load_state_dict(optimizer_state)
+        print("Loaded optimizer state.")
+    else:
+        print("Optimizer state not found, skipping.")
+
+    training_state_path = os.path.join(checkpoint_path, "training_state.pt")
+    if os.path.exists(training_state_path):
+        training_state = torch.load(training_state_path)
+        loaded_rollout_id = training_state.get("iteration", -1)
+        global_step = training_state.get("global_step", 0)
+        print(f"Loaded training state: iteration={loaded_rollout_id}, global_step={global_step}")
+    else:
+        print("Training state not found, skipping.")
+
+    return loaded_rollout_id, global_step
--- a/slime/backends/fsdp_utils/arguments.py
+++ b/slime/backends/fsdp_utils/arguments.py
@@ -31,6 +31,12 @@ class FSDPArgs:
     wandb_project: str = "slime-fsdp"
     wandb_run_name: Optional[str] = None
 
+    # Checkpoint
+    save: Optional[str] = None
+    load: Optional[str] = None
+    save_safe_serialization: bool = True
+    overwrite_checkpoints: bool = False
+
     # Precision
     gradient_checkpointing: bool = False
 


PopSoda2002:
Can you share some test case and result?

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,75 @@
...
+            state_dict = model.state_dict()
+
+        model.save_pretrained(
+            checkpoint_dir,
+            state_dict=state_dict,
+            safe_serialization=args.save_safe_serialization,
+        )
+        tokenizer.save_pretrained(checkpoint_dir)
+
+        torch.save(optimizer.state_dict(), os.path.join(checkpoint_dir, "optimizer.pt"))
```
MrAta: shouldn't we get the full state dict of the optimizer?
something like:
```python
optimizer_state_dict = FSDP.optim_state_dict(self.model, self.optimizer)
```

ppraneth: I’ve made the changes to it . Could you check it and verify if it looks correct?

ppraneth:
> Can you share some test case and result?

I’ve written a test case, but I’m not sure if it’s correct. Since I don’t have access to GPUs, could you please check it and run it?

PopSoda2002:
> Contributor

synced in slack

leng-yue:
Thanks for your great contribution :)

I’d suggest switching to FSDP’s distributed checkpointing, as it could make saving/loading the optimizer and LLM weights noticeably faster and reduce peak memory use. Happy to prototype this if you want.

We can also let FSDP write `safetensors` directly (see the [PyTorch blog on HuggingFace safetensors support](https://pytorch.org/blog/huggingface-safetensors-support-in-pytorch-distributed-checkpointing/)) Example: [link](https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/examples/fsdp_checkpoint_example.py).

PopSoda2002:
> Thanks for your great contribution :)
> 
> I’d suggest switching to FSDP’s distributed checkpointing, as it could make saving/loading the optimizer and LLM weights noticeably faster and reduce peak memory use. Happy to prototype this if you want.
> 
> We can also let FSDP write `safetensors` directly (see the [PyTorch blog on HuggingFace safetensors support](https://pytorch.org/blog/huggingface-safetensors-support-in-pytorch-distributed-checkpointing/)) Example: [link](https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/examples/fsdp_checkpoint_example.py).

Hi thanks for your suggestion! I think we should follow this tutorial and example too, unless there would be OOM for large model with current logic @ppraneth

slime/backends/fsdp_utils/actor.py
```diff
@@ -59,10 +60,14 @@ def init(self, args, role, wandb_run_id, with_ref: bool = False):  # type: ignor
         self.args = args
         torch.manual_seed(args.seed)
 
+        hf_checkpoint = args.load if args.load else args.hf_checkpoint
+        loaded_rollout_id = -1
+        self.global_step = 0
```
PopSoda2002, line 86: What does these 2 variable mean? loaded_rollout_id and self.global_step

leng-yue:
> > Thanks for your great contribution :)
> > I’d suggest switching to FSDP’s distributed checkpointing, as it could make saving/loading the optimizer and LLM weights noticeably faster and reduce peak memory use. Happy to prototype this if you want.
> > We can also let FSDP write `safetensors` directly (see the [PyTorch blog on HuggingFace safetensors support](https://pytorch.org/blog/huggingface-safetensors-support-in-pytorch-distributed-checkpointing/)) Example: [link](https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/examples/fsdp_checkpoint_example.py).
> 
> Hi thanks for your suggestion! I think we should follow this tutorial and example too, unless there would be OOM for large model with current logic @ppraneth

For reference, with the current loading logic a 235B model is likely to OOM on most machines. The bf16 weights alone are ~470 GB (235B × 2 bytes), and with typical 8 GPUs the effective memory can exceed ~3.7 TB.

Once this PR lands, I’m happy to follow up with a change that switches us to FSDP distributed checkpointing (writing `safetensors` directly) to make load/save faster and reduce peak memory.

slime/backends/fsdp_utils/actor.py
ppraneth: loaded_rollout_id:
For a new run: If we are not loading a checkpoint, loaded_rollout_id stays at its default value of -1. The function then returns -1 + 1 = 0, correctly telling the training manager to start with the very first rollout, rollout 0.

When resuming: If we load a checkpoint that was saved after completing rollout number 10, the load_checkpoint function will update loaded_rollout_id to 10. The init function then returns 10 + 1 = 11, correctly telling the manager to resume training from rollout 11.

self.global_step:  tracks how many times the model's weights have actually been updated.

ppraneth:
> > Thanks for your great contribution :)
> > I’d suggest switching to FSDP’s distributed checkpointing, as it could make saving/loading the optimizer and LLM weights noticeably faster and reduce peak memory use. Happy to prototype this if you want.
> > We can also let FSDP write `safetensors` directly (see the [PyTorch blog on HuggingFace safetensors support](https://pytorch.org/blog/huggingface-safetensors-support-in-pytorch-distributed-checkpointing/)) Example: [link](https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/examples/fsdp_checkpoint_example.py).
> 
> Hi thanks for your suggestion! I think we should follow this tutorial and example too, unless there would be OOM for large model with current logic @ppraneth

I will try to implement it

slime/backends/fsdp_utils/actor.py
PopSoda2002: ok, but why not initialize `loaded_rollout_id = 0`?

ppraneth: **Current Way (Starts at -1)**

The code has one simple rule: **the next rollout is always `loaded_rollout_id + 1`**.

* **Starting a New Run**:
    1.  `loaded_rollout_id` starts at **-1**.
    2.  The `init` function returns `-1 + 1 = 0`.
    3.  The training loop starts with **rollout 0**. ✅ **(Correct)**

* **Resuming from a Checkpoint**:
    1.  Let's say the last completed rollout was **rollout 0**. The checkpoint saves this value.
    2.  `load_checkpoint` reads the file and sets `loaded_rollout_id` to **0**.
    3.  The `init` function returns `0 + 1 = 1`.
    4.  The training loop resumes with **rollout 1**. ✅ **(Correct)**

This works perfectly with a single line of code.

 **Alternative Way (If we started at 0)**

If we initialized `loaded_rollout_id` to `0`, we would need two different rules, which makes the code a bit more complex.

* **Starting a New Run**:
    1.  `loaded_rollout_id` starts at **0**.
    2.  The `init` function would have to just return `loaded_rollout_id` (so, `0`).
    3.  The training loop starts with **rollout 0**. ✅ **(Works so far)**

* **Resuming from a Checkpoint (Here's the problem)**:
    1.  The last completed rollout was **rollout 0**.
    2.  `load_checkpoint` would read the file and set `loaded_rollout_id` to **0**.
    3.  Now what? The `init` function can't just return `0` again. It needs to know that this is a *resumed* run and return `1`. The code would need an extra check, like `if (resuming) { return loaded_rollout_id + 1; } else { return loaded_rollout_id; }`.

By using **-1**, we can easily avoided the extra check.

ppraneth:
@PopSoda2002 Can you check the new checkpoint.py 

**Refactored `checkpoint.py`**:
   - Introduced `_flatten_optimizer_state` to flatten the optimizer state dictionary into fully qualified names (FQNs) like `state.<param_name>.<state_key>` and `param_groups.<index>.<key>` for safetensors compatibility.
   - Added `_unflatten_optimizer_state` to reconstruct the nested optimizer state when loading safetensors checkpoints, ensuring compatibility with `set_state_dict`.

Tested with the test_fsdp_save_resume.sh script and works

This new checkpoint can save to safetensors

slime/backends/fsdp_utils/actor.py
PopSoda2002: But why does resuming will return a `+1` while starting not? I mean start should return 0 right? If we load the initial model, it should return 0 not 1 right? I think current way is hard code and tricky, there should be a better way

slime/backends/fsdp_utils/checkpoint.py
PopSoda2002, line 1: I think we can only consider FSDPv2 and `distributed checkpoint`. It should not be so complex and verbose

PopSoda2002 (commented):
Thanks for your working, can you paste a test screenshot in the PR desciption, and I think this PR should not be so complex and verbose

PopSoda2002:
@leng-yue Hi, can you please help review this PR too? I think you are more familiar with this, thanks

slime/backends/fsdp_utils/checkpoint.py
ppraneth: **Changes Made:**

 Now the code explicitly handles three cases:

1. **Resuming from a checkpoint** (has `training_state.pt`): Returns `loaded_iteration + 1` because we want to continue from the next iteration after what was saved
2. **Loading a base model** (no `training_state.pt`): Returns `0` to start fresh training
3. **No `--load` flag**: Returns `0` for fresh training

The key insight is that when we save a checkpoint at iteration 5, that checkpoint contains the model state *after* iteration 5 completes. So when resuming, we should start at iteration 6, not repeat iteration 5.

For loading an initial HuggingFace model (which has no training state), we correctly start from iteration 0.

I also added clear logging for each case so it's obvious what's happening:
- "Resuming training from iteration X" 
- "No training checkpoint found, starting from iteration 0"
- "Starting fresh training from iteration 0"

The `+1` now only happens when actually resuming, not as a trick to fix the default case.

ppraneth:
image

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,203 @@
...
+    """Save a checkpoint using FSDP distributed checkpointing (FSDP v2 only)."""
+    checkpoint_dir = os.path.join(args.save, str(iteration))
+
+    if dist.get_rank() == 0:
+        if os.path.exists(checkpoint_dir):
+            if not args.overwrite_checkpoints:
+                print(
+                    f"WARNING: Checkpoint {checkpoint_dir} exists. Skipping (use --overwrite-checkpoints to overwrite)."
+                )
+                dist.barrier()
```
leng-yue: There’s an issue with the distributed barrier. If `overwrite_checkpoints` is `False`, all ranks except rank 0 call `get_state_dict` and try to write the checkpoint, which can cause the process to hang. A better approach is to check `overwrite_checkpoints` before calling `save_checkpoint` — if it’s `False`, skip the `save_checkpoint` call. Or maybe broadcast `if checkpoint exists` from rank 0.

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,203 @@
...
+
+            # Flatten optimizer for safetensors compatibility
+            flattened_optim = _flatten_optimizer_state(optimizer_state_dict)
+            state_dict["optim"] = flattened_optim
+
+            # Create FQN mapping
+            fqn_to_index = {f"model.{k}": 0 for k in model_state_dict.keys()}
+            fqn_to_index.update({f"optim.{k}": 0 for k in flattened_optim.keys()})
+
+            storage_writer = HuggingFaceStorageWriter(path=checkpoint_dir, fqn_to_index_mapping=fqn_to_index)
```
leng-yue: We may want to save the model and optimizer in separate checkpoints. That makes it much easier to load the model for testing with `from_pretrained`. You can also save or copy the Hugging Face config into the same folder.

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,203 @@
...
+
+            if dist.get_rank() == 0:
+                print("Saving in safetensors format")
+
+        except ImportError:
+            if dist.get_rank() == 0:
+                print("WARNING: safetensors not available, using standard format")
```
leng-yue: Raise an error to avoid silent fail should be better.

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,203 @@
...
+            state_dict = {"model": model_state_dict, "optim": {}}
+            storage_reader = HuggingFaceStorageReader(path=checkpoint_path)
+            dist_cp.load(state_dict=state_dict, storage_reader=storage_reader)
+
+            # Unflatten optimizer state
+            optimizer_state_dict = _unflatten_optimizer_state(state_dict["optim"])
+
+        except ImportError:
+            if dist.get_rank() == 0:
+                print("WARNING: safetensors reader not available, using standard reader")
```
leng-yue: If the `is_safetensors` is True, that means there is only `safetensor` files in the folder, you should raise an error.

slime/backends/fsdp_utils/checkpoint.py
ppraneth: Thanks for catching the barrier hang! I moved the overwrite_checkpoints check to save_model in actor.py, where rank 0 checks existence and broadcasts a should_save flag to all ranks via dist.broadcast. If False, all ranks skip calling save_checkpoint, avoiding get_state_dict and write attempts

slime/backends/fsdp_utils/checkpoint.py
ppraneth: Good point on separate checkpoints for from_pretrained ease. Updated save_checkpoint to save model to model/ subdir (safetensors or .distcp) and optimizer to optimizer/ subdir (standard format), with config.json and tokenizer files in top-level via config.save_pretrained and tokenizer.save_pretrained.

slime/backends/fsdp_utils/checkpoint.py
ppraneth: Yea, silent fallback is risky. Changed the except `ImportError in safetensors` save block to raise `ImportError("Safetensors library required when save_safe_serialization=True")` instead of warning and fallback, ensuring explicit failure if safetensors is missing.

slime/backends/fsdp_utils/checkpoint.py
ppraneth: Updated the except `ImportError` in load's safetensors block to raise` ImportError("Safetensors library required to load safetensors files")`

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -39,29 +28,44 @@ def save_checkpoint(args, iteration, model, optimizer, tokenizer, global_step):
...
+            # Create FQN mappings for model and optimizer
+            model_fqn_to_index = {f"model.{k}": 0 for k in model_state_dict.keys()}
+            optim_fqn_to_index = {f"optim.{k}": 0 for k in flattened_optim.keys()}
 
-            storage_writer = HuggingFaceStorageWriter(path=checkpoint_dir, fqn_to_index_mapping=fqn_to_index)
+            # Save model state dict
+            model_storage_writer = HuggingFaceStorageWriter(
+                path=checkpoint_dir, fqn_to_index_mapping=model_fqn_to_index
+            )
+            dist_cp.save(state_dict={"model": model_state_dict}, storage_writer=model_storage_writer)
```
leng-yue: Do we need a "model" prefix here? This may block `from_pretrained` loading?

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -104,31 +110,44 @@ def load_checkpoint(args, model, optimizer):
...
-        except ImportError:
-            if dist.get_rank() == 0:
-                print("WARNING: safetensors reader not available, using standard reader")
-            state_dict = {"model": model_state_dict, "optim": optimizer_state_dict}
-            storage_reader = dist_cp.FileSystemReader(checkpoint_path)
-            dist_cp.load(state_dict=state_dict, storage_reader=storage_reader)
-            optimizer_state_dict = state_dict["optim"]
+            # Load model state dict
+            model_state_dict = {"model": model_state_dict}
+            model_storage_reader = HuggingFaceStorageReader(path=checkpoint_path)
```
leng-yue: similar here

slime/backends/fsdp_utils/checkpoint.py
leng-yue: Cool!

tests/test1.py
PopSoda2002, line 1: LGTM! But can you remove these tests like tests/test1.py and tests/test2.py, however, you can use these tests to test locally and paste the results in the PR

ppraneth:
image
image

@PopSoda2002 Testing is done

fzyzcjy:
this is pretty useful for long runs

ppraneth:
@zhuzilin Can you check this

fzyzcjy:
@ppraneth hi could you please handle the lint error, and may I know whether it is ready to use (e.g. is it tested and looks good) or still has some ongoing work/fix?

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Get sharded state dicts
+    model_state_dict, optimizer_state_dict = get_state_dict(
+        model, optimizer, options=StateDictOptions(full_state_dict=False, cpu_offload=False)
+    )
+
+    if not optimizer_state_dict or not optimizer_state_dict.get("state", {}):
+        raise ValueError(f"Optimizer state dictionary is empty for iteration {iteration}")
+
+    use_safetensors = getattr(args, "save_safe_serialization", False)
```
fzyzcjy, line 25: tiny nit

```suggestion
    use_safetensors = args.save_safe_serialization
```

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,170 @@
...
+                path=model_subdir, fqn_to_index_mapping={k: 0 for k in model_state_dict.keys()}
+            )
+            dist_cp.save(state_dict=model_state_dict, storage_writer=model_writer)
+        except ImportError as e:
+            raise ImportError(
+                "Safetensors library is required when save_safe_serialization is True, but it is not installed."
+            ) from e
+    else:
+        model_writer = dist_cp.FileSystemWriter(model_subdir)
+        dist_cp.save(state_dict={"model": model_state_dict}, storage_writer=model_writer)
```
fzyzcjy, line 51: tiny nit:

```
if use_safetensors:
  model_writer = HuggingFaceStorageWriter(..)
else:
  model_writer = dist_cp.FileSystemWriter(..)
dist_cp.save(..)
```

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,170 @@
...
+            dist_cp.load(state_dict=model_state_dict, storage_reader=model_storage_reader)
+        except ImportError as e:
+            raise ImportError(
+                "Safetensors library is required to load safetensors checkpoint files, but it is not installed."
+            ) from e
+    else:
+        model_state_dict = {"model": model_state_dict}
+        model_storage_reader = dist_cp.FileSystemReader(model_subdir)
+        dist_cp.load(state_dict=model_state_dict, storage_reader=model_storage_reader)
+        model_state_dict = model_state_dict["model"]
```
fzyzcjy, line 129: (same)

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Load optimizer (always standard format)
+    optim_state_dict = {"optim": optimizer_state_dict}
```
fzyzcjy, line 132: nit: wondering whether we can remove that extra "optim" nested key

slime/backends/fsdp_utils/checkpoint.py
```diff
@@ -0,0 +1,170 @@
...
+            training_state = torch.load(training_state_path, map_location="cpu")
+            loaded_iteration = training_state.get("iteration", -1)
+            global_step = training_state.get("global_step", 0)
+
+        # Broadcast to all ranks
+        state_t = torch.tensor([0, 0], dtype=torch.int64, device="cpu")
+        if dist.get_rank() == 0:
+            state_t[0] = loaded_iteration
+            state_t[1] = global_step
+        dist.broadcast(state_t, src=0)
```
fzyzcjy, line 165: tiny nit: dist.broadcast_object_list

zhuzilin:
close as solved by #633

PR #423, opened 2025-10-04
[doc] fix Multimodal printing bug (type error risk)

lancerts:
1. fix Multimodal printing bug (type error risk)
2. fix doc string


lancerts:
@zhuzilin

PR #427, opened 2025-10-05
Adding tensorboard

zhaochenyang20:


zhaochenyang20:
Also, I added `pkill -9 redis` in the training scripts for quickly restart new experiments.

PR #432, opened 2025-10-06
[Feat]: add pipelined weight update

GeLee-Q:
### Optimize update_weights performance by replacing synchronous ray.get() calls

with a sliding window approach that pipelines multiple update operations.
- Replace per-iteration blocking ray.get() with windowed pipeline execution
- Add configurable pipeline window (size=4) to manage memory and prevent OOM
- Use separate CUDA stream for asynchronous parameter pre-fetching to prevent
rollout-side index errors and illegal memory access issues

image

### DIFF
--- a/slime/backends/megatron_utils/update_weight_utils.py
+++ b/slime/backends/megatron_utils/update_weight_utils.py
@@ -1,9 +1,10 @@
+import collections
 import inspect
 import re
 import socket
 import time
 from argparse import Namespace
-from typing import Any, Iterator, Mapping, Optional, Sequence
+from typing import Any, Iterator, List, Mapping, Optional, Sequence, Tuple
 
 import ray
 import torch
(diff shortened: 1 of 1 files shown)


slime/backends/megatron_utils/update_weight_utils.py
```diff
@@ -375,12 +376,53 @@ def update_weights(self) -> None:
...
+                current_params = next_params
+                current_infos = next_infos
+
+        if rank == self._ipc_gather_src and ray_refs:
+            ray.get(list(ray_refs))
 
         dist.barrier(group=get_gloo_group())
 
-    def _update_bucket_weights_from_tensor(self, param_infos: Sequence[ParamInfo]) -> None:
+    def _update_bucket_weights_from_tensor(
```
zhuzilin: It would be better to rename the `_update_bucket_weights_from_tensor` to perhaps `_gather_bucket_params` since it will no longer update weights after this pr.

slime/backends/megatron_utils/update_weight_utils.py
```diff
@@ -438,33 +480,41 @@ def _update_bucket_weights_from_tensor(self, param_infos: Sequence[ParamInfo]) -
...
+                self._model_update_groups,
+                self.weight_version,
+                self.distributed_rollout_engines,
+                converted_named_tensors,
             )
+            if refs_distributed:
+                all_refs.extend(refs_distributed)
 
-        ray.get(refs)
+        return all_refs if all_refs else None
```
zhuzilin: we can always return `all_refs` here.

slime/backends/megatron_utils/update_weight_utils.py
```diff
@@ -375,12 +376,53 @@ def update_weights(self) -> None:
...
+                    next_params, next_infos = self._update_bucket_weights_from_tensor(
+                        self.param_info_buckets[i + 1]
+                    )
+
+            refs = self._update_converted_params_from_tensor(current_params, current_infos)
+            if refs:
+                if isinstance(refs, list):
+                    ray_refs.extend(refs)
+                else:
+                    ray_refs.append(refs)
```
zhuzilin: if we can always return `all_refs` in `_update_converted_params_from_tensor`, we can simplify the code into:
```python
ray_refs.append(self._update_converted_params_from_tensor(current_params, current_infos))
```
note that we may need to use `append` as the `pipeline_window_size` should be the value for window for number of buckets, and there can be more than one ref for each bucket.

slime/backends/megatron_utils/update_weight_utils.py
```diff
@@ -375,12 +376,53 @@ def update_weights(self) -> None:
...
+                    oldest_ref = ray_refs.popleft()
+                    ray.get(oldest_ref)
+
+            if i + 1 < num_buckets:
+                torch.cuda.current_stream().wait_stream(side_stream)
+                current_params = next_params
+                current_infos = next_infos
+
+        if rank == self._ipc_gather_src and ray_refs:
+            ray.get(list(ray_refs))
```
zhuzilin: after changing the `ray_refs` into `list[list[handles]]`, we need to do a for loop here to sync all the handles.
or, sth like ray.get(sum(ray_refs, [])).

PR #442, opened 2025-10-10
Change trainer backend from environment variable to args

fzyzcjy:


fzyzcjy:
wait a minute, will dump more

PR #444, opened 2025-10-10
[Hardware] AMD - Replace vllm CuMemAllocator dependency with torch_memory_saver

yushengsu-thu:
- Replace vllm CuMemAllocator dependency with torch_memory_saver ([PR](https://github.com/fzyzcjy/torch_memory_saver/pull/52)) on AMD side and ensure slime still works well
- Update the `dockerfile.rocm` and update the dokcer img at [rlsys/slime repo](https://hub.docker.com/r/rlsys/slime/tags)

Screenshot 2025-10-10 at 11 46 37 AM

### DIFF
--- a/slime/backends/megatron_utils/actor.py
+++ b/slime/backends/megatron_utils/actor.py
@@ -9,11 +9,9 @@ from typing import Dict, Optional, Tuple, Union
 import ray
 import torch
 import torch.distributed as dist
-from ray.actor import ActorHandle
-
-from torch_memory_saver import torch_memory_saver
-
 from megatron.core import mpu
+from ray.actor import ActorHandle
+from torch_memory_saver import torch_memory_saver
 from transformers import AutoConfig, AutoTokenizer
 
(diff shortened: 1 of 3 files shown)


yushengsu-thu:
@fzyzcjy @zhuzilin 
Could you please review this PR thanks.

slime/ray/train_actor.py
```diff
@@ -57,17 +57,21 @@ def init(self, args, role, wandb_run_id, with_ref=False):
         args.world_size = dist.get_world_size()
 
         try:
-            import pynvml
+            if torch.version.hip is not None:
```
fzyzcjy, line 60: tiny nit: shall we move if-else outside try

yushengsu-thu: @fzyzcjy 
I leave a comment here: `# will find the corresponding API to implement ROCm version as below`
I will use the same/similar API to implement setting the NUMA affinity. So, it might be good to leave the try outside the if-else. What do you think about this?

PR #446, opened 2025-10-10
fix the sft arguments' default value

YuchenFan48:
When initializing actors, num-rollouts and rollout-num-gpus are used but is set to None by default in SFT setting, leading to illegal operation. I change the default value of rollout-num-gpus to 8 and num-rollouts to 1 which addresses the issue


zhuzilin:
it seems that the issue is caused by wrongly init http client when there is no rollout engine, should have been fixed by commit 529f15af6bfcd4d65268471c14f93bc51093e03d, could you try to pull the latest code and give it another try?

YuchenFan48:
Sure. This is a more elegant solution😄

PR #447, opened 2025-10-10
Move dist.barrier() Outside Loop to Reduce Synchronization Overhead

none0663:
### Summary
This PR moves the dist.barrier() call outside the serialization loop to eliminate unnecessary synchronization overhead while maintaining the same correctness guarantees.
### Problem
The current implementation calls dist.barrier() inside the loop, may result in:

- Excessive synchronization: Barrier invoked world_size times instead of once

- Performance degradation: All processes wait unnecessarily after each iteration

- Unclear synchronization logic: Multiple barrier calls obscure the actual synchronization point


zhuzilin:
the reason for the inner loop is that when training on multiple nodes and the model is saved on a sharing nfs, huggingface transformers will fail on concurrent read/write.

PR #454, opened 2025-10-10
GSPO

### LINKED ISSUE #297: [FSDP] Implement GSPO algorithm
The FSDP backend currently lacks support for the GSPO ([Generalized Second-Order Policy Optimization](https://arxiv.org/abs/2507.18071)) algorithm. This is necessary for MoE training.

ppraneth:
This pull request introduces support for the **Group Sequence Policy Optimization (GSPO)** algorithm within the FSDP backend.  It also closes #297 

This implementation aligns with the principles outlined in the GSPO paper
### **Changes Made**

The core of this implementation is within `slime/backends/fsdp_utils/actor.py`. The key changes include:

1.  **Advantage Estimator Support**:
    * Added `"gspo"` to the list of supported `advantage_estimator` options.
    * The advantage and return calculation now treats rewards on a per-sequence basis for GSPO, which is consistent with the existing GRPO implementation.

2.  **GSPO Policy Loss Calculation**:
    * A new conditional block has been added to the `train` method to handle the GSPO logic.
    * This block calculates the **sequence-level importance ratio (`s_i`)** by taking the exponential of the length-normalized sum of log-probability ratios for each sequence.
    * The final policy loss is computed using the clipped surrogate objective, as described in the GSPO paper.
    * To preserve the computation graph and enable gradient flow, `torch.stack` is used instead of `torch.tensor` when creating new tensors from lists.
    * Tensors are explicitly moved to the correct device to prevent device mismatch errors during loss calculation.

3.  **Numerical Stability**:
    * To prevent division-by-zero errors for empty sequences, I've added `.clamp_min(1)` to the denominator when calculating the sequence-level importance ratio. This is to ensure numerical stability.

### **Testing**

I have successfully tested this implementation using a custom script (`tests/test_gspo.sh`). The script trains a `Qwen3-0.6B` model on the `dapo-math-17k` dataset using the FSDP backend. The training job completed successfully, and the logs confirm that the GSPO loss is being calculated correctly and gradients are flowing as expected.

### DIFF
--- a/slime/backends/fsdp_utils/actor.py
+++ b/slime/backends/fsdp_utils/actor.py
@@ -313,55 +313,50 @@ class FSDPTrainRayActor(TrainRayActor):
             ppo_kl = old_log_probs.to(device=log_probs.device) - log_probs
 
             if self.args.advantage_estimator == "gspo":
-                advantages_seq = torch.stack(
-                    [adv_seq.mean() for adv_seq in torch.split(advantages, response_lengths, dim=0)]
+                log_ratio_splits = torch.split(ppo_kl, response_lengths, dim=0)
+
+                seq_kls = [
+                    ((log_ratio_i * mask_i).sum() / mask_i.sum().clamp_min(1))
+                    for log_ratio_i, mask_i in zip(log_ratio_splits, loss_masks)
+                ]
+
+                ppo_kl_list = []
+                for seq_kl, length in zip(seq_kls, response_lengths):
+                    ppo_kl_list.append(seq_kl.expand(length))
+
+                ppo_kl = torch.cat(ppo_kl_list)
+
+            pg_loss, pg_clipfrac = compute_policy_loss(ppo_kl, advantages, self.args.eps_clip, self.args.eps_clip_high)
+
+            # Apply TIS before sample mean calculation
+            if self.args.use_tis:
+                # Initialize TIS variables
+                tis = None
+                tis_clipfrac = None
+                ois = None
+                # Apply TIS off-policy correction using importance sampling
+                assert all(
+                    "rollout_log_probs" in batch
+                    and isinstance(batch["rollout_log_probs"], torch.Tensor)
+                    and batch["rollout_log_probs"].numel() > 0
+                    for batch in unpacked_batches
+                ), "rollout_log_probs must be provided as non-empty torch.Tensor for TIS"
+
+                rollout_log_probs = torch.cat([batch["rollout_log_probs"] for batch in unpacked_batches], dim=0)
+                rollout_log_probs = rollout_log_probs.to(device=log_probs.device)
+
+                tis = torch.exp(old_log_probs - rollout_log_probs)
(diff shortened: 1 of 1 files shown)


ppraneth:
@zhuzilin Can you please review it?

ppraneth:
[output.txt](https://github.com/user-attachments/files/22855692/output.txt)
This file contains the testing log

slime/backends/fsdp_utils/actor.py
```diff
@@ -304,49 +304,74 @@ def train(self, rollout_id, rollout_data_ref):
...
+                    [
+                        torch.exp((log_ratio_i * mask_i).sum() / mask_i.sum().clamp_min(1))
+                        for log_ratio_i, mask_i in zip(log_ratio, loss_masks)
+                    ]
+                )
+
+                clipped_s_i = torch.clamp(s_i, 1.0 - self.args.eps_clip, 1.0 + self.args.eps_clip)
+                pg_loss = -torch.mean(torch.min(s_i * advantages_seq, clipped_s_i * advantages_seq))
+                pg_clipfrac = torch.mean((s_i != clipped_s_i).float())
+                ppo_kl = sum_of_sample_mean((old_log_probs - log_probs).abs(), response_lengths, loss_masks)
```
zhuzilin: I think that most code here can be shared with the grpo branch. And we only need to  difference is that we need to calculate a new sequence level `ppo_kl`

ppraneth: I have placed the common code above the if else block. Can you review it again

slime/backends/fsdp_utils/actor.py
```diff
@@ -303,50 +303,65 @@ def train(self, rollout_id, rollout_data_ref):
...
-
-                rollout_log_probs = torch.cat([batch["rollout_log_probs"] for batch in unpacked_batches], dim=0)
-                rollout_log_probs = rollout_log_probs.to(device=log_probs.device)
-
-                tis = torch.exp(old_log_probs - rollout_log_probs)
-                ois = (-ppo_kl).exp()
-                tis_clip = torch.clamp(
-                    tis, min=getattr(self.args, "tis_clip_low", 0.1), max=getattr(self.args, "tis_clip", 2.0)
+
+            if self.args.advantage_estimator == "gspo":
```
yitianlian, line 315: I believe we should maintain the GSPO implementation in the FSDP backend to be consistent with the GSPO implementation in the Megatron backend, as this would make it clearer for users. Besides, TIS approach can be applied to both GSPO and GRPO.
We can implement gspo like [https://github.com/THUDM/slime/blob/6d01709727c8710ca397fc494672aae63166aa4e/slime/backends/megatron_utils/loss.py#L282](url)

ppraneth: @yitianlian I have made a couple of changes, can you once again go through changes ?

yitianlian: Thanks! I think this implementation is much clearer ~

PR #458, opened 2025-10-11
Super tiny add logs when kl checker fails

fzyzcjy:


slime/backends/megatron_utils/model.py
```diff
@@ -477,9 +477,9 @@ def train(rollout_id, model, optimizer, opt_param_scheduler, data_iterator, num_
 
             if args.ci_test and not args.ci_disable_kl_checker:
                 if step_id == 0 and "train/ppo_kl" in log_dict and "train/pg_clipfrac" in log_dict:
-                    assert log_dict["train/ppo_kl"] == 0.0 and log_dict["train/pg_clipfrac"] == 0.0
+                    assert log_dict["train/ppo_kl"] == 0.0 and log_dict["train/pg_clipfrac"] == 0.0, f"{log_dict=}"
```
reyoung, line 480: this code will implicitly format log_dict even assert is not failed.

if log_dict contains GPU tensor, it will always implicitly sync device.

Maybe

```
if log_dict["train/ppo_kl"] != 0.0 and log_dict["train/pg_clipfrac"] != 0.0:
    raise ValueError(f"{log_dict=}")
```

is better

:)

fzyzcjy: hmm from my naive understanding the format part will not be called in normal cases. but if you find an extra implicit sync please ping me!

https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-assert_stmt

image

PR #459, opened 2025-10-11
fix: Correctly display sglang_tensor_parallel_size in startup logs

Gao016:
This PR fixes an issue where the sglang_tensor_parallel_size was incorrectly logged as 1 at startup logs, even when a different value was effectively being used via --rollout-num-gpus-per-engine. This misleading log could cause confusion during debug.
For example, if we set the :
SGLANG_ARGS=(
   --rollout-num-gpus-per-engine 8
   --sglang-mem-fraction-static 0.7
   --sglang-cuda-graph-bs 1 2 4 8 $(seq 16 8 256)
)

Before fix, the startup log is：
image
image
which is very confusion when we try different sglang tpsize config through setting rollout-num-gpus-per-engine, it all displays 1.

After fix, the startup log is right:
image
image

### DIFF
--- a/slime/utils/arguments.py
+++ b/slime/utils/arguments.py
@@ -961,14 +961,14 @@ def get_slime_extra_args_provider(add_custom_arguments=None):
                 default=None,
             )
             return parser
-        
+
         def add_sglang_tp_size():
             temp_parser = argparse.ArgumentParser(add_help=False)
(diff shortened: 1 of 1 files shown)


zhuzilin:
wow, thank you for this! can you fix the lint with the pre-commit hook?

Gao016:
> wow, thank you for this! can you fix the lint with the pre-commit hook?

ok~ i have fix the lint with the pre-commit hook. thanks~

PR #476, opened 2025-10-13
[DO NOT MERGE] check ci

fzyzcjy:
I see IMA in ci script and thus want to firstly check whether this also appear here, main does not run ci thus need a pr


fzyzcjy:
hmm seems it exists
https://github.com/THUDM/slime/pull/475

image

PR #484, opened 2025-10-13
[Doc ]Include type annotation and doc string for update_weight_utils.py

lancerts:


slime/backends/fsdp_utils/update_weight_utils.py
```diff
@@ -17,24 +20,23 @@
...
 
 try:
-    from sglang.srt.model_executor.model_runner import FlattenedTensorBucket
+    from sglang.srt.model_executor.model_runner import FlattenedTensorBucket  # type: ignore[import]
 
     use_flattened_tensor_bucket = True
 except ImportError:
     use_flattened_tensor_bucket = False
 
-from slime.utils.memory_utils import clear_memory
```
yitianlian, line 26: I  believe this function should not be deleted, as it is used on lines 126 and 306.

lancerts: This function has been imported already in line 16, hence duplicated import.

PR #485, opened 2025-10-13
[Doc ]Include type annotation and doc string for data.py

lancerts:


slime/utils/types.py
```diff
@@ -52,7 +52,9 @@ def get_reward_value(self, args) -> float:
 class ParamInfo:
     name: str
     dtype: torch.dtype
-    shape: torch.Size
-    attrs: dict
-    size: int
-    src_rank: int
+
```
yitianlian, line 59: I think these attributions should not be deleted.

slime/backends/megatron_utils/data.py
```diff
@@ -398,7 +484,7 @@ def log_perf_data(rollout_id, args):
             if not args.wandb_always_use_train_step
             else rollout_id * args.rollout_batch_size * args.n_samples_per_prompt // args.global_batch_size
         )
-        if args.use_wandb:
+        if args.use_wandb and wandb is not None:
```
yitianlian: Why do we need `wandb is not None`? Wandb is just a package, not a variable.

lancerts: Nice catch, corrected.

slime/utils/types.py
lancerts: Yeah, accidental removal, corrected.

PR #486, opened 2025-10-14
Super tiny code cleanup

fzyzcjy:
found when reading training loss computations


slime/backends/megatron_utils/loss.py
```diff
@@ -105,8 +106,6 @@ def get_values(
     unconcat_tokens: list[torch.Tensor],
     total_lengths: list[int],
     response_lengths: list[int],
-    with_entropy: bool = False,
-    non_loss_data: bool = True,
```
zhuzilin, line 155: The `non_loss_data` is required by megatron (https://github.com/NVIDIA/Megatron-LM/blob/48406695c4efcf1026a7ed70bb390793918dd97b/megatron/core/pipeline_parallel/schedules.py#L246).

PR #507, opened 2025-10-15
[Doc] Update annotation in types.py

lancerts:
Following https://peps.python.org/pep-0585/?utm_source


zhuzilin:
hmm... I think that rewrite the code strictly to the typing requirements from pep is not necessary and may have negative influence toward the readability. I'll close all the pr with only type annotation change.

PR #510, opened 2025-10-15
[Feat] Support retool multi-turn partial rollout

gramesh-amd:
Adding partial rollout support for retool example.

Summary of the changes:
- Added reconstruct_loss_masks(response, tokenizer):
  - Rebuilds loss masks from a saved response when resuming partial rollout
  - Marks <interpreter> blocks as non-trainable (0), model output as trainable (1)
  - Used when resuming without saved metadata or when the response length and the loss mask lengths dont match (this rarely happens in training when resuming partial rollout)
- Added count_tool_turns(response) - this just counts completed tool calls by counting </interpreter> tags (helpful to know which turn we are in when resuming partial rollout if not already in metadata)
- Add a check to make sure we don't sample outside model's max context length (args.rollout_max_context_len and a SAFETY_MARGIN to keep track of how much context we have left to generate)
- Some metadata to track the formatted prompt, loss_masks, response so far etc for partial rollout resumption

Tests:
image
image

- Unfortunately both non partial rollout example already provided and the partial rollout support ive added have both bad training results. I have tested both of these runs in AMD MI325x and might need to also test on nvidia gpus and potentially needs a hyperparameter tune to get better results.

- I've manually verified the partial rollout rollouts and all stages + resumption seems fine.

@yushengsu-thu @zyzshishui Can you guys review this and let me know what we can do to make it better?

Thanks

### DIFF
--- a/examples/retool/generate_with_retool.py
+++ b/examples/retool/generate_with_retool.py
@@ -336,6 +336,10 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
         tool_call_count = 0
         start_turn = 0
 
+
+    output = None
+
+    
     for turn in range(start_turn, TOOL_CONFIGS["max_turns"]):
         current_length = len(prompt_tokens_ids) + len(response_token_ids)
         remaining_context = MAX_CONTEXT_LENGTH - current_length - SAFETY_MARGIN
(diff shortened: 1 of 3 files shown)


yushengsu-thu (changes requested):
@gramesh-amd 
1. Could you have another individual script for AMD, since it might have different environment variables for use?
2. Only add the partial rollout args in the old script - .*sh

gramesh-amd:
image
image
There is step count mismatch between amd and nvidia runs (see x axis but both runs have been run for similar time. Although b200 get oom error much faster so i probably need to change the hyperparams to get a longer run)

Zhangmj0621:
@gramesh-amd @yushengsu-thu Hi, about this pr, I have a worry about that what will happen if partial-rollout happens when one sample was processing tool call? In that case, I bother if the sample may have some unpredictable behaviour.

PR #520, opened 2025-10-16
[FSDP][ready for review] Reduce peak memory required when gathering log-probs

tyler-romero:
Small change that will reduce the memory requirements of gathering logprobs in the actor. This saves memory by using torch.compile to fuse the log-softmax and gather operations to avoid materializing the full logprob tensor.

See benchmark here: https://github.com/allenai/open-instruct/pull/584#issuecomment-2686851911


zhuzilin (approved):
Thanks a lot!

PR #523, opened 2025-10-16
[DOC] Update type annotation based on PEP guide

lancerts:


zhuzilin:
hmm... I think that rewrite the code strictly to the typing requirements from pep is not necessary and may have negative influence toward the readability. I'll close all the pr with only type annotation change.

PR #530, opened 2025-10-17
Fix error when training is resumed

fzyzcjy:


fzyzcjy:
close since hear this conflicts w/ some code

PR #545, opened 2025-10-21
Support advantage normalization and loss aggregation options

sam571128:
Title. 

Inspired by this paper: https://arxiv.org/pdf/2510.13786. By adding advantage normalization & loss aggregation options, this provides us two benefits:
1. Easier support of new algorithms. We can support DAPO, CISPO etc by changing these options as the loss of most RL algorithms are built on top of these different options.
2. Easier configuration of trainings based on existing algorithms.

I might have done something incorrectly as I did not have the full context of how some of these were implemented.

Let me know what you think, and what will need to be changed!

### DIFF
--- a/slime/backends/megatron_utils/loss.py
+++ b/slime/backends/megatron_utils/loss.py
@@ -709,14 +709,19 @@ def loss_function(
         loss * num_microbatches / args.global_batch_size * mpu.get_data_parallel_world_size(with_context_parallel=True)
     )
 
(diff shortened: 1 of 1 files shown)


sam571128:
Well. Actually prompt level loss aggregation here doesn't work due to dynamic batch size. I can see why it wasn't supported before. Have to set `micro_batch_size` to be a multiple of `n_samples_per_prompt`

PR #551, opened 2025-10-21
init for drgrpo

maocheng23:
Implementation of DrGRPO


sam571128:
I think this is already supported when you use `--calculate-per-token-loss` and `--disable-grpo-std-normalization`.

PR #552, opened 2025-10-22
[docker] trim megatron and sglang patch

zhuzilin:


shuowoshishui:
bug

``` 
  def sleep(self) -> None:
      assert self.args.offload_train

      clear_memory()
      print_memory("before offload model")
      destroy_process_groups()

      torch_memory_saver.pause()

      print_memory("after offload model")
```
The colocate mode uses the latest Slime code and an error occurred here. Single-node and multi-node experiments were conducted using H20 and L40 respectively. This error occurs whenever the `destroy_process_groups()` function is used; commenting it out allows it to run normally.I'd like to ask about the specific purpose of this code. Why not use the previous mp.destroy_process_groups()?Is it possible to provide a WeChat group for communication?Thanks

PR #557, opened 2025-10-23
Fix logprob does not handle temperature

fzyzcjy:


fzyzcjy:
(maybe cur logprobs need to handle temperature, should do later)

PR #560, opened 2025-10-23
Add kimi-k2

Gao016:
As the title says.


zhuzilin:
wow, this is great! thank you so much! could you share your running script in the PR? I don't have resource to check such a big model at the moment.

Gao016:
> Thank you so much!
Thanks~~ 
i am testing the K2 running script this week and close to success, I'll upload it as soon as possible~~

PR #561, opened 2025-10-23
Fix: rollback pipeline update weights because fp8 rollout bug when rollout tp > 1

lilei199908:


fzyzcjy:
I am confused what is this... @zhuzilin @lilei199908 

1. why save the same thing in if-else branch
2. why do not save rollout manager (this will be buggy)

image

fzyzcjy:
tried a naive fix https://github.com/THUDM/slime/pull/564

PR #569, opened 2025-10-24
Split args.offload to train and rollout

fzyzcjy:


slime/utils/arguments.py
```diff
@@ -76,11 +76,20 @@ def add_cluster_arguments(parser):
                 ),
             )
             parser.add_argument(
-                "--offload",
```
zhuzilin, line 79: can we still leave the old `--offload`, where it will enable both `offload_train` and `offload_rollout` for backward compatible?

fzyzcjy: sure

PR #584, opened 2025-10-25
[FSDP] Fix ref model loading

Hecate0821:


zhuzilin:
sorry it seems that the fsdp codebase went through huge refactor recently, could you check if the current impl still have the issue here?

zyzshishui:
> sorry it seems that the fsdp codebase went through huge refactor recently, could you check if the current impl still have the issue here?

Still exist. Move to https://github.com/THUDM/slime/pull/656

Hecate0821:
> sorry it seems that the fsdp codebase went through huge refactor recently, could you check if the current impl still have the issue here?

It still exists, we will work on the new PR yuzhen mentioned above

PR #585, opened 2025-10-25
Support multi task evaluation

zyzshishui:


slime/utils/eval_config.py
```diff
@@ -0,0 +1,117 @@
...
+import copy
+from dataclasses import dataclass, field
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple
+
+
+_EMPTY_VALUES = (None, [], {})
+
+
+@dataclass
+class EvalDatasetConfig:
```
fzyzcjy, line 12: if we use dataclass, what about Pydantic

slime/rollout/rm_hub/gpqa.py
fzyzcjy, line 1: qq: where is this judge come from (good to be from some de-facto libraries to be extra safe)

slime/utils/eval_config.py
zyzshishui: Added in https://github.com/THUDM/slime/pull/595

slime/rollout/rm_hub/gpqa.py
zyzshishui: From vibe coding + human checking (⁄ ⁄•⁄ω⁄•⁄ ⁄)
But the logic is quite common. I have used the script to evaluate on Qwen3-4B (with thinking). The officially reported score is 55.9, so I think that's reliable.
截屏2025-10-25 下午3 42 04

fzyzcjy: oh i see. I personally suggest copy-pasting from e.g. lm-eval, NemoSkills, etc, to be extra safe, but it is not a big issue for this multi-choice question

PR #603, opened 2025-10-26
Compile rope while preserving true on policy

fzyzcjy:


slime/backends/fsdp_utils/actor.py
```diff
@@ -53,8 +53,10 @@ class FSDPTrainRayActor(TrainRayActor):
     def init(self, args: Namespace, role: str, wandb_run_id: str, with_ref: bool = False) -> int:  # type: ignore[override]
         super().init(args, role, wandb_run_id, with_ref)
 
+        # TODO extract to function
```
fzyzcjy, line 56: fyi I will extract it separately to avoid making large diffs

PR #604, opened 2025-10-27
[Feature] Add sglang server metrics in wandb logging

yitianlian:


slime/utils/sglang_metric.py
```diff
@@ -0,0 +1,697 @@
+import time
```
zhuzilin, line 1: I wonder if it is possible for us to use `x_stats_open_metrics_endpoints` in wandb for the each rollout engines?

fzyzcjy:
hmm this seems to be already done

(w/ https://github.com/sgl-project/sglang/pull/11456 in review progress)

slime/utils/sglang_metric.py
fzyzcjy: seems this is exactly what I have done -> https://github.com/THUDM/slime/blob/83a3c964d68eb7f5cc8b2d05ef165d4ede4df813/slime/utils/wandb_utils.py#L102

fzyzcjy: (about why I do not do it for each of the rollout engines but instead let sgl-router do aggregation: iirc doing that will make it hard to be displayed in wandb nicely)

yitianlian: I discuss with [slin1237](https://github.com/slin1237) regarding logging metrics, and he mentioned that he hopes to merge the `engine_metrics `and `list_workers `endpoints into a single one. He also mentioned that he has reached out to you regarding the PR for this change.  @fzyzcjy 

I’m writing this because there’s an urgent need to profile and adjust some parameter settings. Once the SGL router supports metrics aggregation, I’ll remove this workaround.

fzyzcjy: > He also mentioned that he has reached out to you regarding the PR for this change

seems I pinged him (https://sgl-fru7574.slack.com/archives/C09E1U4LL6Q/p1761389280211429) and he has not checked yet

> I’m writing this because there’s an urgent need to profile and adjust some parameter settings. Once the SGL router supports metrics aggregation, I’ll remove this workaround.

I see, then it LGTM since it is temporary. just a nit: it would be great to somehow extract the logic (e.g. a few simple functions) from rollout.py to minimize code diff in that file

PR #611, opened 2025-10-27
Fix non thinking model deepscaler score bug 

guapisolo:
When rollout non-thinking model, like Qwen-4B-2507., the score is always 0.
This is because the response does not have `<think>` prefix or `###Response` prefix.


zhuzilin:
this is not really a bug. the origin deepscaler rm contains a format reward.

guapisolo:
> this is not really a bug. the origin deepscaler rm contains a format reward.

Understand. I will close it.

PR #621, opened 2025-10-28
update `slime.utils.compute_approx_kl` for clarity

Arist12:
- Fix docstring: remove the non-existent `action_mask`,  and add `kl_loss_type`
- Remove redundant `log_ratio` calculation in k2 case
- Use descriptive variable name (`approx_kl`) instead of reusing `log_ratio`
- Fix grammar in comments
- Simplify negation syntax in k3


zhuzilin:
close as most part of the pr is rewriting the existing code without linting.

PR #625, opened 2025-10-29
Implement benchmark_func for performance measurement

zlH518:
Added a benchmarking function to measure execution time and performance metrics.


zhuzilin:
hmm... this is not normally needed in a rl framework because it is rare to test a part of the model (which should be done within megatron or sglang), close for now, and maybe reopen when we have such need.

PR #632, opened 2025-10-29
Refactor and add timers in training

fzyzcjy:


slime/utils/timer.py
```diff
@@ -15,12 +17,16 @@ def __init__(self):
...
+        if torch.distributed.get_rank() == 0:
+            print(f"Timer {name} start")
 
     def end(self, name):
         assert name in self.start_time, f"Timer {name} not started."
         elapsed_time = time() - self.start_time[name]
         self.add(name, elapsed_time)
         del self.start_time[name]
+        if torch.distributed.get_rank() == 0:
+            print(f"Timer {name} end (elapsed: {elapsed_time:.1f}s)")
```
zhuzilin, line 29: could you remove those 2 prints? I'm afraid this will cause too much print in the log.

fzyzcjy: I personally find it pretty useful since it "splits" the log into sections, and this only introduce ~10 lines of log while one step may have ~100 (or ~1000) lines of log from sglang rollout, but if you like I can remove it

zhuzilin: ok, make sense.

PR #642, opened 2025-10-30
Avoiding multiple ranks write into debug train data file concurrently

YuchenFan48:
When enabling ``--save-debug-train-data``, all ranks write into the ``.pt`` file concurrently, resulting in ``PytorchStreamReader failed reading file data/3: invalid header or archive...``. 
We first write data of different ranks into different temp files and merge them into one to avoid the problem.

### DIFF
--- a/slime/backends/megatron_utils/actor.py
+++ b/slime/backends/megatron_utils/actor.py
@@ -334,14 +334,10 @@ class MegatronTrainRayActor(TrainRayActor):
             self.prof.after_actor_train_step(rollout_id=rollout_id)
         
         if (path_template := self.args.save_debug_train_data) is not None:
-            import time
             from pathlib import Path
         
             rank = torch.distributed.get_rank() if torch.distributed.is_initialized() else 0
-        
-            path_str = path_template.format(rollout_id=rollout_id, rank=rank)
-            path = Path(path_str)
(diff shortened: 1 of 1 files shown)


fzyzcjy:
wondering maybe just write to different files

YuchenFan48:
Though a little time is wasted, merging into one file results in a clear file structure. Otherwise, one step will store ``num_ranks`` files🤔

fzyzcjy:
> results in a clear file structure

hmm but multi files seems common, e.g. a megatron/fsdp ckpt

YuchenFan48:
Maybe use a file structure like below?
```
save_path/
├── rollout_id_0/
│   ├── rank_0.pt
│   ├── rank_1.pt
│   └── ...
├── rollout_id_1/
│   └── ...
└── ...
```

fzyzcjy:
then maybe change this line to

```
        args.save_debug_train_data = f"{args.dump_details}/train_data/{{rollout_id}}/rank_{{rank}}.pt"
```

etc

image

fzyzcjy:
I still do not think we should merge the files... :/ especially when we have the folder structure

YuchenFan48:
Done. Maybe you can check again.

PR #667, opened 2025-11-01
Support megatron backend in 4b demo script

fzyzcjy:


fzyzcjy:
this may be delayed from merging since we need 4b fsdp script for the true on policy now

fzyzcjy:
verified locally this can run (for the true on policy), no regression

PR #672, opened 2025-11-02
[FSDP] Add an example script to run 4B demo with fsdp

Zhuohao-Li:
As titled, provide an aligned style script to run Qwen3-4B w/ FSDP under `scripts`

I suggest to make more detailed doc for FSDP user regarding the optimization args.

### DIFF
--- a/scripts/run-qwen3-4B-fsdp.sh
+++ b/scripts/run-qwen3-4B-fsdp.sh
@@ -1,0 +1,134 @@
+#!/bin/bash
+
+# for rerun the task
+pkill -9 sglang
+sleep 3
+ray stop --force
+pkill -9 ray
+pkill -9 python
+sleep 3
+pkill -9 ray
+pkill -9 python
+
+set -ex
+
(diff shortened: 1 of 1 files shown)


Zhuohao-Li:
image
image
@zhuzilin

fzyzcjy:
qq: what is your truncate ratio vs time? iirc we deprecated to use 8k rollout + qwen3-4b b/c it is somehow learning to speak shorter instead of learning math...

also could you please also paste the eval aime accuracy

Zhuohao-Li:
> qq: what is your truncate ratio vs time? iirc we deprecated to use 8k rollout + qwen3-4b b/c it is somehow learning to speak shorter instead of learning math...
> 
> also could you please also paste the eval aime accuracy

yes, it is indeed shorter...

image

it is better to use 32768 rollout intuitively, maybe we should also update `scripts/run-qwen3-4B.sh` w/ megatron, and use extra eval bench (gpqa/math) since aime-24 has limited data

PR #683, opened 2025-11-04
Fix SGLang Router Endpoint

sam571128:
After rebasing internal Slime with the current slime, the rollout no longer works. After diagnosing the issue, I realized these two endpoints are deprecated, and switching to the new endpoint works.


fzyzcjy:
yes this is needed, o/w i encounter

```
(RolloutManager pid=519445) 2025-11-08 10:07:09 ERROR sglang_router_rs::response: src/middleware.rs:256: request failed with server error
(RolloutManager pid=519445) Error: Server error '503 Service Unavailable' for url 'http://10.78.210.151:4674/generate'
(RolloutManager pid=519445) For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503, retrying... (attempt 60/60, url=http://10.78.210.151:4674/generate, response=No available workers (all circuits open or unhealthy))
```

but maybe we should do a if-else, e.g. import sgl router and check its version, and only enable if it is new

zhuzilin (approved):

PR #684, opened 2025-11-04
[perf]:Optimize Dataset Loading/Filtering with Multiprocessing

none0663:
## Overview
This PR optimizes the slow tokenization filtering in dataset loading by using multiprocessing to bypass Python's GIL.
https://github.com/THUDM/slime/pull/404

## Changes
- Add `_batch_tokenize_and_filter()` helper for parallel processing
- Use `ProcessPoolExecutor` for datasets with >100 samples
- Add optional `num_workers` parameter (defaults to `cpu_count() // 2`, max 8)
- Preserve original sequential logic for small datasets and multimodal cases

## Performance
- **Large datasets (>100 samples)**: 2-4x speedup
- **Small datasets**: No impact (uses original logic)
- **Multimodal datasets**: No impact (uses original logic)

## Key Points
- Uses **multiprocessing** (not multi-threading) to avoid GIL
- Minimal code changes, fully backward compatible
- No breaking changes

### DIFF
--- a/slime/utils/data.py
+++ b/slime/utils/data.py
@@ -6,6 +6,7 @@ import numpy as np
 import pandas as pd
 import ray
 import torch.distributed as dist
+from tqdm import tqdm
 
 from slime.utils.types import Sample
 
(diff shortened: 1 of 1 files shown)


zhuzilin:
hmm... I'll close this as the tokenizer is most of the time not the performance bottleneck and if we need to optimize this part, maybe a lazy tokenizer would be better.

PR #501, opened 2025-10-15
[FSDP, VLM] feat: add vlm training for FSDP

nanjiangwill:
Goal: Support VLM training on slime with FSDP

TODO
- [X] model support
- - [X] add support for `class AutoModelForVision2Seq/AutoModelForImageTextToText`
- - [X] support `Qwen2.5-VL/Qwen3-VL`(we just use default hf) new-feature TBD  @windowsXP-beta
- [ ] support dataset - refer to examples in `examples`
- - [X] for single turn -> [geo3k](https://huggingface.co/datasets/hiyouga/geometry3k), how verl did it https://github.com/volcengine/verl/pull/2014 @coding-famer @nanjiangwill 
- - [ ] for multi turn -> [deepeyes](https://arxiv.org/abs/2505.14362), how verl did it https://github.com/volcengine/verl/pull/2398
- [X] single turn VLM training
- - [X] support image in prompt/data, image processing, data pipeline @coding-famer @nanjiangwill 
- - [X] dataset properly handle multimodal info @coding-famer 
- - [X] rollout properly handle multimodal info @coding-famer @nanjiangwill 
- - [X] training(fsdp) properly handle multimodal info @nanjiangwill [this](https://github.com/volcengine/verl/blob/main/verl/workers/actor/dp_actor.py#L117)
- - [X] able to train on geo3k
- [ ] multi turn VLM training @minleminzui
- - [X] data pipeline for new images generated during the rollout stage

PR #687, opened 2025-11-05
Tiny refactor multimodal data input preprocessing

coding-famer:
Fix the multimodal preprocessing logic to
1. fix incorrect condition logic like https://github.com/THUDM/slime/blob/16b39196a1c2c831b0f0b210d554e4cba33ef2da/slime/rollout/sglang_rollout.py#L100
2. support conversational format input
3. use `<image>` mark for image position

to support multi-turn multi-image input


coding-famer:
Move to #501

PR #688, opened 2025-11-05
Add Search-R1 example for Importance Sampling

ChangyiYang:
# Add Search-R1 Example with Local Index and TIS Support

## Changes

1. **Added local dense retriever support in `generate_with_search.py`**
   - Supports both local search (FAISS index) and Google search backends
   - Added setup instructions in README for local retriever (requires ~132GB disk space)

2. **Added Trajectory Importance Sampling (TIS) support in `generate_with_search.py`**
   - Demonstrated TIS usage in `run_qwen2.5_3B.sh` with commented examples
   - Collects log probabilities while maintaining token/logp alignment

3. **More IS Metrics**
   - Add chi2 related metric in mis.py

## References

- Original Search-R1: https://github.com/PeterGriffinJin/Search-R1

### DIFF
--- a/examples/train_infer_mismatch_helper/mis.yaml
+++ b/examples/train_infer_mismatch_helper/mis.yaml
@@ -16,11 +16,11 @@ mis_mode: "mask"
 # For mask or clip mode, the lower bound of the IS weights.
 # For truncate mode, it will not be used.
 # If not set, it will be set to 1.0 / mis_upper_bound
-# Geo should use 0.9999
+# For Geometry level, the lower bound should be 0.9999
 mis_lower_bound: 0.5
 
 # For truncate, mask, or clip mode, the upper bound of the IS weights
-# Geo should use 1.0001
+# For Geometry level, the upper bound should be 1.0001
 mis_upper_bound: 2.0
 
 # Per-token veto threshold. If any token ratio < this, zero the entire sequence weight, the sequences won't have gradient


examples/search-r1/generate_with_local_search_logp.py
```diff
@@ -0,0 +1,203 @@
...
+    pattern = r"<(search|answer)>(.*?)</\1>"
+    match = re.search(pattern, prediction, re.DOTALL)
+    if match:
+        content = match.group(2).strip()  # Return only the content inside the tags
+        action = match.group(1)
+    else:
+        content = ""
+        action = None
+
+    return action, content
```
zhaochenyang20: which one is actually being used? Also, adding comments here to say that why we can not do postprocessing on the token level for the rollout log probs.

examples/search-r1/generate_with_search.py
```diff
@@ -12,7 +12,7 @@
 SEARCH_R1_CONFIGS = {
     "max_turns": 3,
     "topk": 3,
-    "google_api_key": "YOUR_API_KEY",  # Replace with your actual API key
+    "google_api_key": "6f32c0e10507ce84f69751cda90614374bccf35a",  # Replace with your actual API key
```
zhaochenyang20: where to get the key? Adding guidance in the docs:

https://eidgn0bxdh.feishu.cn/wiki/OwJ6wcuSZiPdEqkL28ccVfG2n3c

examples/train_infer_mismatch_helper/mis.yaml
```diff
@@ -5,7 +5,7 @@ use_tis: true
 # token: per-token
 # sequence: product over tokens
 # geometric: geometric mean
-mis_level: "token"
+mis_level: "sequence"
```
zhaochenyang20: I think we shall not change this

zhaochenyang20:
To me, this PR only adds the scripts to run the search engine and give scripts right? No modification on the SearchR1 side. If so, it's good to go.

examples/search-r1/generate_with_search.py
```diff
@@ -12,7 +12,7 @@
 SEARCH_R1_CONFIGS = {
```
Zhuohao-Li, line 14: I think it's better to merge the current three generation logic functions in a unified file for simplicity.
we can expose config like this to set local/serp search or w/wo IS

ChangyiYang:
> To me, this PR only adds the scripts to run the search engine and give scripts right? No modification on the SearchR1 side. If so, it's good to go.

Not exactly — I also removed the post-processing part for SGLang responses in order to collect raw logprobs.
Other than that, there are indeed no further changes. ✅

examples/search-r1/generate_with_search.py
ChangyiYang: Great advice! Will do!

zhaochenyang20: Hey, please do as Zhuohao's suggestions. Then I will merge

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #689, opened 2025-11-05
perf: Add pipelined weight update via --pipeline-update flag

GeLee-Q:
Performance Benchmark: K2-FP8 Weight Update Time

Update Strategy | 0.5B Parameters | 2B Parameters | 4B Parameters | 6B Parameters | 8B Parameters
-- | -- | -- | -- | -- | --
Pipelining Enabled | 518 s | 191 s | Unstable* | – | –
Pipelining Disabled | 921 s | – | 239 s | 219 s | OOM

These tests were performed in collaboration with @Gao016 @lilei199908

### DIFF
--- a/slime/backends/megatron_utils/update_weight_utils.py
+++ b/slime/backends/megatron_utils/update_weight_utils.py
@@ -503,7 +503,6 @@ class UpdateWeightFromTensor:
                     )
             for handle in handles:
                 handle.wait()
-        torch.cuda.synchronize()
 
         # broadcast params across ep ranks
         if ep_size > 1:


GeLee-Q:
@lilei199908

slime/backends/megatron_utils/update_weight_utils.py
```diff
@@ -464,6 +503,7 @@ def _gather_bucket_params(
                     )
             for handle in handles:
                 handle.wait()
+        torch.cuda.synchronize()
```
lilei199908: It seems that this line is not necessary?

GeLee-Q: This was added to address the precision issues that occur during fp8 tp2 rollouts.

slime/backends/megatron_utils/update_weight_utils.py
```diff
@@ -482,6 +522,7 @@ def _gather_bucket_params(
...
```
lilei199908, line 524: I tested that only this change can address the issues during FP8 TP2 rollouts, and the previous sync is not needed.

GeLee-Q: Got it. I'll get on it.

PR #694, opened 2025-11-06
Add kimi-k2-instruct running script

Gao016:
As the title says, I run it for 40 steps, and the raw_rewards is shown in the following figure:
image
This work is done in collaboration with @GeLee-Q @yefei12 and @yzlnew

### DIFF
--- a/scripts/run-kimi-k2-Instruct.sh
+++ b/scripts/run-kimi-k2-Instruct.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+
+# for rerun the task
+pkill -9 sglang
+sleep 3
+ray stop --force
+pkill -9 ray
+pkill -9 python
+sleep 3
+pkill -9 ray
+pkill -9 python
+
+set -ex
+
+# will prevent ray from buffering stdout/stderr
+export PYTHONBUFFERED=16
+
+NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
+if [ "$NVLINK_COUNT" -gt 0 ]; then
+    HAS_NVLINK=1
+else
+    HAS_NVLINK=0
+fi
+echo "HAS_NVLINK: $HAS_NVLINK (detected $NVLINK_COUNT NVLink references)"
+
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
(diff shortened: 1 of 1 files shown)


zhuzilin (approved):

fzyzcjy:
Hi, I would appreciate it if figures about eval accuracy could be shown.

Also I am curious, why the initial raw rewards is so low, is it b/c it truncates a lot, and the model is only learning to speak briefly, or something else

Gao016:
> Hi, I would appreciate it if figures about eval accuracy could be shown.
> 
> Also I am curious, why the initial raw rewards is so low, is it b/c it truncates a lot, and the model is only learning to speak briefly, or something else

Hi~ thanks for your question.
It's probably because this reward is totally OOD, which explains why the scores were low at the beginning.
The rollout/truncated is show in the following figure:
image
Regrettably, the eval logs were not preserved. We are now running the task again and then supply the corresponding figures.
@GeLee-Q @yefei12 @yzlnew

fzyzcjy:
I see, looking forward to it!

> It's probably because this reward is totally OOD, which explains why the scores were low at the beginning.

hmm why is math reward OOD - in my naive understanding models are usually pretty good at math. maybe it is the format that is ood (?) then looks like we are testing the format-learning

Gao016:
> I see, looking forward to it!
> 
> > It's probably because this reward is totally OOD, which explains why the scores were low at the beginning.
> 
> hmm why is math reward OOD - in my naive understanding models are usually pretty good at math. maybe it is the format that is ood (?) then looks like we are testing the format-learning

thanks~ emm, we have also compared the raw reward scores of K2 with Moonlight-16B-A3B and DeepSeek-R1. Interestingly, the 16B model began with a high initial score, while DeepSeek-R1 exhibited a trend similar to K2, starting low and improving slowly. The comparison figure with 16 steps is provided below.
image
We are now conducting further research into this phenomenon and hope to share new findings for discussion.thanks~😄
@GeLee-Q @yefei12 @yzlnew

fzyzcjy:
looking forward to further results. I would personally suggest to use `--dump-details` and have a look at the rollout samples manually (potentially via https://github.com/THUDM/slime/blob/main/slime/utils/debug_utils/display_debug_rollout_data.py)

Gao016:
> looking forward to further results. I would personally suggest to use `--dump-details` and have a look at the rollout samples manually (potentially via https://github.com/THUDM/slime/blob/main/slime/utils/debug_utils/display_debug_rollout_data.py)

ok~thanks, we will try it

fzyzcjy:
you are welcome

PR #695, opened 2025-11-06
fix bug on gspo

zhuzilin:


zhuzilin:
using this pr to test if ci can pass...

PR #696, opened 2025-11-06
Optimize data.py for efficient data loading

ppraneth:
The previous implementation of `read_file` used `pandas.read_json` and `pandas.read_parquet`. This approach loaded the **entire file** into system memory *before* processing, leading to two major issues:
1.  **Massive Memory Usage:** A 10.7GB JSONL file could consume over 94GB of RAM.
2.  **Slow Startup Time:** The program would hang while it spent minutes loading the full dataset.

### Key Changes
To fix this, the `read_file` function was completely rewritten to use a streaming approach, and the `pandas` dependency was removed.

1.  **Streaming `.jsonl`:** Instead of `pd.read_json`, files are now read line-by-line using a generator. This keeps memory usage minimal (megabytes instead of gigabytes).
2.  **Streaming `.parquet`:** Instead of `pd.read_parquet`, files are now read in batches using `pyarrow.parquet.ParquetFile.iter_batches`. This streams data from the file efficiently.
3.  **Dependency Removal:** `pandas` is no longer a dependency for `slime/utils/data.py`.
4.  **Robustness:** Added error handling for malformed lines in `.jsonl` files.

### Benchmark Results
The new streaming implementation was benchmarked against the old `pandas` method on large public datasets (16M+ rows).

**JSONL (Datasource: `TIGER-Lab/Fineweb-Instruct`)**
| Metric | Old (pandas) | New (Streaming) | Improvement |
| :--- | :--- | :--- | :--- |
| **Peak Memory** | 94,663.65 MB | **0.11 MB** | **99.999% less** |
| **Total Time** | 1111.82 sec | **172.28 sec** | **6.5x faster** |

**Parquet (Datasource: `Trimness8/reddit_dataset_145`)**
| Metric | Old (pandas) | New (Streaming) | Improvement |
| :--- | :--- | :--- | :--- |
| **Peak Memory** | 17,300.94 MB | **87.91 MB** | **99.5% less** |
| **Total Time** | 1174.19 sec | **307.45 sec** | **3.8x faster** |

### DIFF
--- a/slime/utils/data.py
+++ b/slime/utils/data.py
@@ -78,17 +78,16 @@ def _parse_generalized_path(s: str):
     return s, None
 
 
-def _should_skip_prompt(
-    prompt, tokenizer, processor, metadata, max_length, apply_chat_template, apply_chat_template_kwargs
-):
+def _should_skip_prompt(formatted_prompt: str, tokenizer, processor, max_length, multimodal_inputs=None):
     if max_length is None:
         return False
 
-    from slime.utils.processing_utils import prepare_model_inputs
+    if processor:
+        processor_output = processor(text=formatted_prompt, **multimodal_inputs)
+        input_ids = processor_output["input_ids"][0]
+    else:
+        input_ids = tokenizer.encode(formatted_prompt, add_special_tokens=False)
 
-    input_ids, _ = prepare_model_inputs(
-        prompt, tokenizer, processor, metadata, apply_chat_template, apply_chat_template_kwargs
-    )
     return len(input_ids) > max_length
 
 
@@ -169,6 +168,7 @@ class Dataset:
             prompt = _build_messages(data, prompt_key, as_conversation, multimodal_keys)
 
             metadata = data.get(metadata_key) or {}
+            tools = None
             if tool_key is not None and tool_key in data:
                 tools = data[tool_key]
                 if isinstance(tools, str):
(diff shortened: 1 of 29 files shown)


ppraneth:
cc @zhuzilin

ppraneth:
@yitianlian can you check this pr?

slime/utils/data.py
```diff
@@ -17,26 +26,52 @@
...
     else:
         raise ValueError(f"Unsupported file format: {path}. Supported formats are .jsonl and .parquet.")
 
     if row_slice is not None:
-        logger.info(f"read_file path={path} slice {len(df)=} rows into {row_slice=}")
-        df = df.iloc[row_slice]
 
-    for _, row in df.iterrows():
-        yield row.to_dict()
+        print(f"read_file path={path} applying slice {row_slice=}")
```
yitianlian: You can change it to logger.info(xxx)

yitianlian:
Hi, the format error might need to be fixed~

slime/utils/data.py
ppraneth: fixed

yitianlian:
Maybe you should run the pre-commit~

ppraneth:
> Maybe you should run the pre-commit~

Done

ppraneth:
Also @yitianlian should i remove the test files?

yitianlian:
Yes, I think these files may not be necessary.

ppraneth:
> Yes, I think these files may not be necessary.

Done!

PR #701, opened 2025-11-08
[WIP] Support SimpleTir

zyzshishui:
To be verified
data: https://huggingface.co/datasets/zyzshishui0627/simpletir
script: examples/simpletir/run_simpletir.sh

### Clone slime
``` bash
git clone https://github.com/zyzshishui/slime.git
git checkout -b simpletir origin/simpletir
```
### Download Data&Model
``` bash
huggingface-cli download --repo-type dataset zyzshishui0627/simpletir --local-dir /root/datasets/simpletir
huggingface-cli download Qwen/Qwen2.5-3B --local-dir /root/Qwen2.5-3B
huggingface-cli download zyzshishui0627/Qwen2.5-3B_torch_dist --local-dir /root/Qwen2.5-3B_torch_dist
```

### Run Script (4 GPU)
``` bash
nohup bash examples/simpletir/run_simpletir.sh > nohup.out 
```

### DIFF
--- a/scripts/run-kimi-k2-Instruct.sh
+++ b/scripts/run-kimi-k2-Instruct.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+
+# for rerun the task
+pkill -9 sglang
+sleep 3
+ray stop --force
+pkill -9 ray
+pkill -9 python
+sleep 3
+pkill -9 ray
+pkill -9 python
+
+set -ex
+
+# will prevent ray from buffering stdout/stderr
+export PYTHONBUFFERED=16
+
+NVLINK_COUNT=$(nvidia-smi topo -m 2>/dev/null | grep -o 'NV[0-9][0-9]*' | wc -l)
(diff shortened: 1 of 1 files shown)


AIDefender:
Hi @zyzshishui, thanks for your effort in supporting SimpleTIR! Have you tested the receipe and check the performance? If not, I would be happy to help. I went through the code and find several mismatches with the original SimpleTIR:
1. The prompt enables the model to call the `final_answer` function in a code block to submit the answer. To do this, SimpleTIR explicitly adds such function to the code block:

```python
if self.config.append_final_answer_func:
   code = (
"""
def final_answer(result):
    print(f"\\\\boxed{{{result}}}")
"""
    + code
)
```
Otherwise, you may just delete the prompt relevant to the `final_answer` function.
2. SimpleTIR executes the first code block and truncates all contents after the first code block. In this receipe it seems that the last code block is executed.
3. SimpleTIR set KL Divergence to 0.

PR #710, opened 2025-11-12
Format code and fix missing arguments

zyzshishui:

PR #711, opened 2025-11-12
Doc of true on policy done 3 week ago

fzyzcjy:
doc of this -> https://github.com/THUDM/slime/pull/566

some text is copied from the doc draft in https://github.com/THUDM/slime/pull/710

thanks @zyzshishui to verify the latest main again:

image


scripts/run_qwen3_4b_fsdp.py
```diff
@@ -18,7 +18,7 @@ class ScriptArgs:
     num_gpus_per_node: int = 8
     hardware: Literal["H100"] = "H100"
     extra_args: str = ""
-    multi_eval: bool = True
+    multi_eval: bool = False
```
fzyzcjy: for simplicity i make multi eval non-default. but that does not harm and is completely unrelated (just to save a bit of time for users who use default flag), since it is just evaling one or multi benchmark.

zhaochenyang20:
Could we tell the users how to turn on true onpolicy on other scripts, like:

"please turn on xxx in xxx`.

zhaochenyang20:
also, pre commit failed

fzyzcjy:
(synced offline)

PR #730, opened 2025-11-14
[Feature] Tiny fix for wandb run id

yitianlian:
It might be a better approach to assign the wandb_run_id in the args to the actual run_id, eliminating the need to explicitly pass the wandb_run_id to each function. Instead, we can directly access it from args.

If the proposal is approved, I will update the function parameter to reflect the change for wandb_run_id.

@fzyzcjy

### DIFF
--- a/tools/convert_hf_to_torch_dist.py
+++ b/tools/convert_hf_to_torch_dist.py
@@ -84,7 +84,7 @@ def main():
         world_size=world_size,
         rank=global_rank,
         device_id=torch.device(f"cuda:{local_rank}"),
-        )
+    )
     args = get_args()
     init(args)
     model = get_model(get_model_provider_func(args), ModelType.encoder_or_decoder, wrap_with_ddp=False)


fzyzcjy:
To be honest, I personally favor an immutable `args` object, but yes today it is already highly mutable (and adding this runid into args does not hurt much since it will never harm any real logic)...

yitianlian:
Thanks for the reply~ I will refactor some of the code related to wandb_run_id.

yitianlian:
I have already finished it. :)

train_async.py
```diff
@@ -9,14 +9,14 @@ def train(args):
     assert not args.colocate, "Colocation is not supported for async training."
     # allocate the GPUs
     pgs = create_placement_groups(args)
-    wandb_run_id = init_wandb_primary(args)
+    args.wandb_run_id = init_wandb_primary(args)
```
fzyzcjy: nit: shall we do 

```
    init_wandb_primary(args)
```

and set args.wandb_run_id inside it, assuming we follow the "args is mutable" approach, given that train script is written by user and we want users to do as little as possible

fzyzcjy:
btw could you please merge from latest main where i updated the ci

fzyzcjy:
LGTM, feel free to merge (and I want to refactor a bit about logging and printing which will be good to be based on this to avoid conflicts)

PR #738, opened 2025-11-14
Super tiny fix typo

fzyzcjy:


WindowsXp-Beta:
Addressed review comments in 955c3581:

- **Shell injection (#1, #2)**: Fixed — using `shlex.quote()` for directory paths in both `du` and `find` commands instead of manual single-quote wrapping.
- **Bare dict nit (#3)**: `dict[str, int]` is already typed — the guideline prohibits untyped `dict`, not typed generics. No change needed.
- **Fragile .get() chain (#4)**: This pattern is copied from `monitor_stale_jobs.py` (the established convention). No change needed.

Issue #774, opened 2025-11-16
[Bug][FSDP] ppo_kl should be zero under onpolicy mode


PopSoda2002:
Currently the ppo_kl does not be zero when using ref model under on policy mode, while not using ref model and kl loss, it is fine:

<img width="689" height="836" alt="Image" src="https://github.com/user-attachments/assets/5efd5fd6-e289-4179-ad08-856308d9e6ef" />

PR #780, opened 2025-11-16
[FSDP]Fix ppo_kl 

Hecate0821:
Fix: https://github.com/THUDM/slime/issues/774
Load ref model separately, but offload to CPU, only load to GPU when needed.
In the PR: https://github.com/THUDM/slime/pull/829
@PopSoda2002 Added a cpu update in the restore phase, make the ppo_kl to 0 as expected. Which proves that our `update_gpu_param_dict` do cause some numerical drifting. After update the "drifted" weight back to CPU, we eliminate the ppo_kl.
To address this drifting issue from the beginning, we can just load the ref model as a separate model ,but only move it to GPU when computing logp for reference. In this way, we eliminate the complex weight swapping and does not introduce extra GPU memory cost.
image


PopSoda2002:
Nice work!

Zhuohao-Li:
LGTM, very solid work!

PR #802, opened 2025-11-17
Add backward compatiblility for sgl-router

zhuzilin:


slime/backends/sglang_utils/sglang_engine.py
```diff
@@ -132,10 +134,15 @@ def _init_normal(self, server_args_dict):
         logger.info(f"Launch HttpServerEngineAdapter at: {self.server_host}:{self.server_port}")
         self.process = launch_server_process(ServerArgs(**server_args_dict))
         if self.node_rank == 0 and self.router_ip and self.router_port:
-            requests.post(
-                f"http://{self.router_ip}:{self.router_port}/workers",
-                json={"url": f"http://{self.server_host}:{self.server_port}"},
-            )
+            if parse(sglang_router.__version__) <= parse("0.2.1"):
```
fzyzcjy, line 137: maybe we also need to consider slime router (if using slime router then regard as old)

Issue #830, opened 2025-11-19
True on policy doesn't work on the latest docker image + main(695a68c)


WindowsXp-Beta:
We tried to reproduce the true on policy RL, i.e. strict 0 logprob diff between sglang inference and FSDP trainer but failed using the latest slime image and main branch(commit: 695a68c). Specifically, the mean and max logprob diff we got are
```
mean: 0.00044722
max: 0.0231
```

Thanks to @guapisolo, we know that latest slime image + slime at 62e2c47 can produce 0 logdiff. There must be some regression from that commit until now.

cc @fzyzcjy

fzyzcjy:
Yes I think so, must be some regression. Could you please help check a bit where it regresses? btw there will be CI to protect this (cc @zhuzilin).

guapisolo:
> Yes I think so, must be some regression. Could you please help check a bit where it regresses? btw there will be CI to protect this (cc [@zhuzilin](https://github.com/zhuzilin)).

I am working on that.

guapisolo:
slimerl/slime:latest + main
Cannot reproduce true on policy

slimerl/slime:latest + 62e2c47367fd9546b530d1ce968e8b3619e51802
Can reproduce true on policy 

slimerl/slime:20251113-v1 + main
Environment error.

slimerl/slime:20251113-v1 + 62e2c47367fd9546b530d1ce968e8b3619e51802
Can reproduce true on policy 

I've been bisecting the problem in the latest docker.

guapisolo:
Emm.. The problem is caused by this commit 6d3b33fc98d0609ba4ed424d7fb5f8be8955da9b. This PR https://github.com/THUDM/slime/pull/467. Could we call @PopSoda2002 to help fix this issue?

PopSoda2002:
Sorry for introducing this, I will help to take a look

guapisolo:
Doesn't matter, let do this together

PopSoda2002:
Solved by PR: https://github.com/THUDM/slime/pull/833, thanks everyone

zhuzilin:
Thanks everyone for this!

PR #833, opened 2025-11-19
[FSDP][BugFix] Avoid autocast for computing log prob for true on policy

PopSoda2002:
Solve issue: https://github.com/THUDM/slime/issues/830#issuecomment-3550209190

image

Thanks so much for @guapisolo 's help!


guapisolo:
@fzyzcjy  Plz review.

PopSoda2002:
> LGTM.
> 
> btw someone mentions we should not use autocast since we are using fsdp. if anyone is interested feel free to do that update (and update sgl's dtype details accordingly)

I can raise an issue first

PopSoda2002:
Issue: https://github.com/THUDM/slime/issues/834

PR #851, opened 2025-11-20
Make router config and data padding configurable, improve FSDP actor code structure

lancerts:
• Extract true_on_policy_mode setup into dedicated method in FSDP actor
• Add --slime-router-timeout and --slime-router-max-connections arguments
• Add --data-pad-size-multiplier argument (default: 128)
• Update get_batch() to use pad_multiplier parameter
• Remove hardcoded values and TODO comments
• Maintain backward compatibility with sensible defaults

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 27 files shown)


lancerts:
cc @zhuzilin ty

lancerts:
cc @yitianlian

yitianlian (approved):
Sorry for the late reply. This pr looks good to me~

PR #861, opened 2025-11-21
[FSDP] only use move mode for fsdp backend and refactor weight updation

zhuzilin:
In the original fsdp backend we reused torch memory saver (tms) offloading from the Megatron backend. This makes it hard to access parameters during weight updates and forces us to introduce extra “param info buckets” as a workaround.

However, tms is only necessary in the Megatron backend because the distributed optimizer makes `model.cpu()` offloading tricky, which is not the case for the fsdp backend.

This PR switches fsdp back to the original offload_train_mode/move flow and keeps tms only for Megatron, resulting in a much cleaner weight update implementation for fsdp.

### DIFF
--- a/slime/backends/fsdp_utils/update_weight_utils.py
+++ b/slime/backends/fsdp_utils/update_weight_utils.py
@@ -66,6 +66,8 @@ class UpdateWeight(abc.ABC):
         if bucket:
             self.wait_and_update_bucket_weights(bucket)
             del bucket
+            bucket = []
+            bucket_size = 0
 
     def wait_and_update_bucket_weights(self, bucket):
         bucket = [(name, param.wait()) if hasattr(param, "wait") else (name, param) for name, param in bucket]


slime/backends/fsdp_utils/update_weight_utils.py
```diff
@@ -1,117 +1,89 @@
...
+        rollout_engines: Sequence[ActorHandle],
+        rollout_engine_lock: ActorHandle | None,
+    ) -> None:
+        pass
 
-class UpdateWeightFromTensor:
+    def update_weights(self) -> None:
+        bucket = []
+        bucket_size = 0
+        for name, param in self.model.state_dict().items():
```
Hecate0821, line 48: This all gather might block here. Should we use something like ` params_list = list(self.model.named_parameters())`?

slime/backends/fsdp_utils/update_weight_utils.py
```diff
@@ -1,117 +1,89 @@
...
+                param = param.redistribute(
+                    placements=[Replicate()] * param.device_mesh.ndim,
+                    async_op=True,
+                ).to_local()
+            bucket.append((name, param))
+            bucket_size += param_size
+
+        if bucket:
+            self.wait_and_update_bucket_weights(bucket)
+            del bucket
```
PopSoda2002, line 68: should we make `bucket_size=0` here? Otherwise, it may have some bugs in the future?

PopSoda2002 (commented):
Nice work!

slime/backends/fsdp_utils/update_weight_utils.py
zhuzilin: fixed.

slime/backends/fsdp_utils/update_weight_utils.py
zhuzilin: It seems that for fsdp2, `model.state_dict()` won't trigger all gather.

Hecate0821: Discussed offline, this will not perform all gather

PR #871, opened 2025-11-24
Add readable data logging

guapisolo:
`--save-debug-rollout-data` is designed for fastest replay. But we still need an easy-to-use data rollout feature for debugging, so I introduce `--save-readable-rollout-data` to help.

image

### DIFF
--- a/slime/ray/rollout.py
+++ b/slime/ray/rollout.py
@@ -185,12 +185,14 @@ class RolloutManager:
         if (path_template := self.args.save_readable_rollout_data) is not None:
             import json
 
-            num_samples = self.args.save_readable_rollout_data_limit
-
             def save_data(path, samples):
                 logger.info(f"Save readable rollout data to {path}")
                 path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


slime/ray/rollout.py
```diff
@@ -179,6 +181,34 @@ def _save_debug_rollout_data(self, data, rollout_id, evaluation: bool):
...
+
+            def save_data(path, samples):
+                logger.info(f"Save readable rollout data to {path}")
+                path.parent.mkdir(parents=True, exist_ok=True)
+                if self.args.save_readable_rollout_data_limit is not None:
+                    num_samples = min(self.args.save_readable_rollout_data_limit, len(samples))
+                    output_data = samples[:num_samples]
+                else:
+                    output_data = samples
+                excluded_fields = {"tokens", "rollout_log_probs", "loss_mask", "rollout_routed_experts"}
```
yitianlian, line 196: I think we can use whether a value can be serialized with json.dump to automatically filter the data, allowing us to support more key fields in the future.

PR #976, opened 2025-11-30
Add strands-agents example 

Lawhy:
Add **slime x strands-agents** example to showcase how to connect modern agentic training framework with agentic scaffolding framework. This example is adapted from the `examples/retool`.

### DIFF
--- a/examples/strands-agents/generate_with_strands.py
+++ b/examples/strands-agents/generate_with_strands.py
@@ -1,19 +1,13 @@
 import logging
 
-from camel.interpreters import SubprocessInterpreter
 import openai
+import wandb
+from camel.interpreters import SubprocessInterpreter
 from strands import Agent, tool
 from strands.models.openai import OpenAIModel
-from strands.types.exceptions import (
-    ContextWindowOverflowException,
-    EventLoopException,
-    MaxTokensReachedException,
-)
-import wandb
+from strands.types.exceptions import ContextWindowOverflowException, EventLoopException, MaxTokensReachedException
 
-from slime.rollout.rm_hub.math_dapo_utils import (
-    compute_score as math_dapo_compute_score,
-)
+from slime.rollout.rm_hub.math_dapo_utils import compute_score as math_dapo_compute_score
 from slime.rollout.sglang_rollout import GenerateState
 from slime.utils.types import Sample
 
--- a/examples/strands-agents/requirements.txt
+++ b/examples/strands-agents/requirements.txt
@@ -1,3 +1,3 @@
+camel-ai
 strands-agents
 strands-agents-tools
-camel-ai


examples/strands-agents/generate_with_strands.py
```diff
@@ -0,0 +1,276 @@
...
+
+    # Set sample attributes and some debug information
+    sample.tokens = prompt_tokens_ids + response_token_ids
+    sample.response_length = len(response_token_ids)
+    sample.response = response_text
+    sample.loss_mask = loss_masks
+    sample.payload_text = prompt_text + response_text
+    sample.payload_has_system = True
+    sample.payload_has_tools = len(agent.tool_names) > 0
+    sample.tool_call_count = [message["role"] == "tool" for message in trajectory].count(True)
```
yitianlian: I believe some of these keys are unnecessary, and the sample class does not support this attribute.

examples/strands-agents/generate_with_strands.py
```diff
@@ -0,0 +1,276 @@
...
+    if wandb.run is not None:
+        wandb.log(
+            {
+                "debug/response_length": sample.response_length,
+                "debug/available_tools": len(agent.tool_names),
+                "debug/tool_calls": sample.tool_call_count,
+                "debug/num_messages": len(trajectory),
+                "debug/truncated": sample.status == Sample.Status.TRUNCATED,
+            }
+        )
```
yitianlian, line 228: I think this log should check whether the x-axis is ‘rollout/step’, ‘step’, or ‘none’?

examples/strands-agents/generate_with_strands.py
```diff
@@ -0,0 +1,276 @@
...
+    return sample
+
+
+async def reward_func(args, sample, **kwargs):
+    """Tool call reward function using math_dapo as primary reward model"""
+    if not isinstance(sample, Sample):
+        raise TypeError("Sample must be an instance of Sample class.")
+
+    # Extract information from sample
+    solution_str = sample.payload_text
```
yitianlian: You can use the sample.response.

examples/strands-agents/README.md
```diff
@@ -0,0 +1,56 @@
+# Slime x Strands-Agents
+
+This is a running example that connects the [Strands-Agents](https://github.com/strands-agents/sdk-python) agent scaffolding framework with Slime for RL training.
+
+## Install Dependencies
+
+1. Pull the `slimerl/slime:latest` image and enter it
+2. Goes to slime folder: `cd /root/slime`
```
yitianlian: Add a clone command?

yitianlian (commented):
I’ve given a few small suggestions.

examples/strands-agents/README.md
Lawhy: I was thinking `/root/slime` is already there in the image? Do we still need to clone?

examples/strands-agents/generate_with_strands.py
Lawhy: done

examples/strands-agents/generate_with_strands.py
Lawhy: checked, default to step

examples/strands-agents/generate_with_strands.py
Lawhy: removed; only kept tool_call_count

PR #990, opened 2025-12-01
Fix per token loss scale and add e2e ci

zhuzilin:


.github/workflows/pr-test.yml
```diff
@@ -44,7 +44,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        info: [{"num_gpus": 8, "test_file": "test_quick_start_glm4_9B.py"}, {"num_gpus": 8, "test_file": "test_qwen3_30B_A3B.py"}, {"num_gpus": 8, "test_file": "test_qwen3_4B_ppo.py"}, {"num_gpus": 8, "test_file": "test_moonlight_16B_A3B.py"}, {"num_gpus": 2, "test_file": "test_qwen3_4B_fsdp_true_on_policy.py"}]
+        info: [
```
fzyzcjy: nit: need to update .j2

PR #998, opened 2025-12-01
Fixed bug in checking max_length for SFT #997

### LINKED ISSUE #997: data.py length check breaks for SFT with message list prompts
## Bug Description

Commit `79fd101` ("fix(data): correct length filtering from character to token level") introduced a regression that breaks SFT training with message list prompts.

The change moved `tokenizer.encode(prompt)` outside the `if not multimodal_keys:` check, causing it to run unconditionally on all prompts, including list-of-dict message prompts used in SFT.

## Reproduction

1. Prepare an SFT dataset with `messages` column containing list of message dicts:
   ```python
   {"messages": [{"role": "user", "content": "..."}, {"role": "assistant", "content": "..."}]}
   ```

2. Run SFT training following `examples/formal_math/single_round/run_sft.py` pattern:
   ```bash
   --rollout-function-path slime.rollout.sft_rollout.generate_rollout
   --input-key messages
   ```

3. Error:
   ```
   TypeError: TextEncodeInput must be Union[TextInputSequence, Tuple[InputSequence, InputSequence]]
   ```

Surya-Gunukula:
## Summary

Add `isinstance(prompt, str)` check to skip token-level length filtering for list prompts (SFT messages).

## Changes

```python
# Before
if max_length is not None:
    raw_prompt_ids = tokenizer.encode(prompt, ...)
    if not multimodal_keys:
        ...

# After
if max_length is not None and not multimodal_keys and isinstance(prompt, str):
    raw_prompt_ids = tokenizer.encode(prompt, ...)
    ...
```

Fixes #997

### DIFF
--- a/slime/utils/data.py
+++ b/slime/utils/data.py
@@ -98,11 +98,11 @@ class Dataset:
                 prompt = prompt_content
 
             # TODO: this is slow.
-            if max_length is not None:
+            # Only check token length for string prompts, skip for multimodal and message lists (SFT)
+            if max_length is not None and not multimodal_keys and isinstance(prompt, str):
                 raw_prompt_ids = tokenizer.encode(prompt, add_special_tokens=False)
-                if not multimodal_keys:
-                    if len(raw_prompt_ids) > max_length:
-                        continue
+                if len(raw_prompt_ids) > max_length:
+                    continue
 
             self.origin_samples.append(
                 Sample(


tengxiaoliu:
Came across the same issue, this looks good.

Lawhy:
Came across the same issue, waiting for update.

Also, I found in https://github.com/THUDM/slime/blob/main/slime/utils/data.py#L101 `max_length` is always not `None` even if I didn't set anything.

Surya-Gunukula:
> Came across the same issue, waiting for update.
> 
> Also, I found in https://github.com/THUDM/slime/blob/main/slime/utils/data.py#L101 `max_length` is always not `None` even if I didn't set anything.

It looks like in arguments.py, --rollout-max-response-len defaults to 1024 if no value is provided. Then, if --rollout-max-context-len is not specified, it also defaults to the same value as rollout-max-response-len. Several other length-related arguments are similarly backfilled using --rollout-max-response-len when their explicit values are None.

Because of that, max_length will never actually be None, which means the current if max_length is None check is effectively dead code.

Nice catch!

PR #999, opened 2025-12-02
[Feature] Add off-policy sequence masking algorithm proposed in DeepSeek v3.2

yitianlian:


slime/utils/arguments.py
```diff
@@ -811,6 +811,18 @@ def add_algo_arguments(parser):
                 default=False,
                 help="The rollout routing replay technique from https://arxiv.org/abs/2510.11370",
             )
+            parser.add_argument(
+                "--enable-opsm",
```
zhuzilin: please rename to `--use-xxx` to align the naming to other flags. and I wonder if we can have an additional `--use-off-policy-sequence-mask` because the origin paper does not mention `opsm` abbreviation.

also please add the dpsk v3.2 paper to the help.

slime/backends/fsdp_utils/actor.py
```diff
@@ -591,24 +592,34 @@ def _train_step(self, packed_batch, reported_accum, mbs_id, grad_accum):
         response_lengths = [batch["response_lengths"] for batch in unpacked_batches]
 
         advantages = advantages.to(device=log_probs.device)
-        ppo_kl = old_log_probs.to(device=log_probs.device) - log_probs
+        old_log_probs = old_log_probs.to(device=log_probs.device)
+        ppo_kl = old_log_probs - log_probs
+
+        opsm_mask = None
+        opsm_clipfrac_num = 0
+        if getattr(self.args, "enable_opsm", False):
```
zhuzilin:
```suggestion
        if self.args.use_opsm:
```

slime/backends/fsdp_utils/actor.py
```diff
@@ -591,24 +592,34 @@ def _train_step(self, packed_batch, reported_accum, mbs_id, grad_accum):
         response_lengths = [batch["response_lengths"] for batch in unpacked_batches]
 
         advantages = advantages.to(device=log_probs.device)
-        ppo_kl = old_log_probs.to(device=log_probs.device) - log_probs
+        old_log_probs = old_log_probs.to(device=log_probs.device)
+        ppo_kl = old_log_probs - log_probs
+
+        opsm_mask = None
+        opsm_clipfrac_num = 0
```
zhuzilin: please move these 2 values inside the `if getattr(self.args, "enable_opsm", False):` and always use `enable_opsm` as the flag to check if opsm is enabled.

slime/backends/megatron_utils/loss.py
```diff
@@ -403,27 +405,55 @@ def policy_loss_function(
...
-        ]
-        full_old_log_probs = [
-            all_gather_with_cp(old_log_prob, total_length, response_length)
-            for old_log_prob, total_length, response_length in zip(
-                old_log_probs, total_lengths, response_lengths, strict=False
-            )
-        ]
+    # Pre-gather log probs if needed by OPSM or GSPO to avoid duplicate gathering
+    cp_size = mpu.get_context_parallel_world_size()
+    need_full_log_probs = (args.enable_opsm or args.advantage_estimator == "gspo") and cp_size > 1
```
zhuzilin:
```suggestion
    need_full_log_probs = (args.enable_opsm or args.advantage_estimator == "gspo")
```
it seems that the code below has deal with cp == 1.

slime/backends/megatron_utils/loss.py
```diff
@@ -403,27 +405,55 @@ def policy_loss_function(
...
+        full_log_probs = (
+            [
+                all_gather_with_cp(log_prob, total_length, response_length)
+                for log_prob, total_length, response_length in zip(
+                    log_probs, total_lengths, response_lengths, strict=False
+                )
+            ]
+            if cp_size > 1
+            else log_probs
+        )
```
zhuzilin: there seems to be early return in `all_gather_with_cp` we can always run all_gather_with cp here.

slime/backends/megatron_utils/loss.py
```diff
@@ -403,27 +405,55 @@ def policy_loss_function(
...
-        loss_masks = batch["loss_masks"]
-        ppo_kl = [
-            ((old_logprob - log_prob) * loss_mask).sum() / torch.clamp_min(loss_mask.sum(), 1)
-            for log_prob, old_logprob, loss_mask in zip(full_log_probs, full_old_log_probs, loss_masks, strict=False)
-        ]
-        ppo_kl = [kl.expand_as(log_prob) for kl, log_prob in zip(ppo_kl, log_probs, strict=False)]
-        ppo_kl = torch.cat(ppo_kl, dim=0)
+    # Compute OPSM mask if enabled
+    opsm_mask = None
+    opsm_clipfrac_num = 0
```
zhuzilin: same as the comment in fsdp, please move into the `args.enable_opsm`

slime/backends/megatron_utils/loss.py
```diff
@@ -433,6 +463,9 @@ def policy_loss_function(
```
zhuzilin, line 451: when running with opsm and without gspo, will the shape of ppo_kl different from opsm_mask?

slime/utils/ppo_utils.py
```diff
@@ -45,6 +47,85 @@ def compute_approx_kl(
...
+
+    for full_log_prob, full_old_log_prob, advantage, loss_mask, local_log_prob in zip(
+        full_log_probs, full_old_log_probs, advantages, loss_masks, local_log_probs, strict=False
+    ):
+        # Calculate sequence-level KL
+        seq_kl = ((full_old_log_prob - full_log_prob) * loss_mask).sum() / torch.clamp_min(loss_mask.sum(), 1)
+
+        # Calculate sequence-level advantage (mean of advantage values)
+        # For GRPO, advantage is constant across the sequence, so mean == any element
+        seq_advantage = advantage.mean()
```
zhuzilin: hmm... it seems that from the dpsk v3.2 paper, we dont need to have per-sequence adv but per token adv.

slime/utils/ppo_utils.py
```diff
@@ -45,6 +47,85 @@ def compute_approx_kl(
...
+
+        # Create sequence-level mask: 0 if (seq_adv < 0 and seq_kl > delta), else 1
+        # This mask applies to the entire sequence
+        condition = (seq_advantage < 0) & (seq_kl > args.opsm_delta)
+        mask_chunk = torch.where(condition, torch.zeros_like(local_log_prob), torch.ones_like(local_log_prob))
+        opsm_clipfrac_num += condition.int().item()
```
zhuzilin: please don't do `.item()` during the loop, it will trigger gpu -> cpu copy and synchronization.

zhuzilin (approved):

PR #1005, opened 2025-12-02
Fix evaluation parameter parsing

guapisolo:
This PR do:

1. Use `MISSING` instead of `None` to make sure dataset yaml config can actively set `null` to avoid overwritten by args. 
2. Move evaluation default param parsing from `sglang_rollout.py` `eval_rollout_single_dataset()` to `eval_config.py` `build_eval_dataset_configs()`. 
3. Fix the parameter name inconsistent between `prompt_key` and `input_key`.
4. Change `EvalDatasetConfig` to `dataclass` type.
5. Add error message when eval dataset path is not found.
6. Remove `min_new_tokens` field.
7. Remove `stop` and `stop_token_ids`.


slime/utils/data.py
```diff
@@ -21,6 +22,9 @@
 def read_file(path):
     path, row_slice = _parse_generalized_path(path)
 
+    if not os.path.exists(path):
```
guapisolo, line 25: This message help debug

PR #1006, opened 2025-12-02
Fix checkpoint conversion from megatron to hf format

raunak-agarwal:
Just need to change the import in `tools/convert_torch_dist_to_hf.py`


yitianlian:
Have fixed it in https://github.com/THUDM/slime/pull/1003. Thanks~

PR #1007, opened 2025-12-02
Fix bare except clause and remove redundant computation in ppo_utils

lancerts:
## Summary
Fixed two code quality issues in the codebase:

1. **Bare except clause in tool_sandbox.py**: Replaced bare `except:` with `except Exception:` 
2. **Redundant computation in ppo_utils.py**: Removed duplicate `log_ratio` calculation in the k2 branch that was already computed earlier


lancerts:
cc @yitianlian

yitianlian (approved):
LGTM！

PR #1025, opened 2025-12-04
Support Tau2 bench evalutation

guapisolo:
Running more tests.


MaxwellJryao:
Hi, are you also doing the model finetuning?

guapisolo:
> Hi, are you also doing the model finetuning?

Nope. This only works for evaluation.

zhuzilin:
close with the same reason as #1156

PR #1027, opened 2025-12-04
refactor: extract duplicated checkpoint interval logic into reusable helper

lancerts:
## Summary
Refactors duplicated eval and save interval checking logic by extracting it into a reusable helper function.

## Changes
- Added `should_run_periodic_action()` helper to `slime/utils/misc.py`
- Updated `train.py` to use helper for both eval and save interval checks
- Updated `train_async.py` to use helper for both eval and save interval checks
- Removed TODO comment about duplicated eval logic

### DIFF
--- a/slime/utils/misc.py
+++ b/slime/utils/misc.py
@@ -67,13 +67,13 @@ def get_free_port(start_port=10000, consecutive=1):
     return port
 
 
-def should_run_checkpoint_action(
+def should_run_periodic_action(
     rollout_id: int,
     interval: int | None,
     num_rollout_per_epoch: int | None = None,
 ) -> bool:
     """
-    Return True when a periodic checkpoint action (eval/save) should run.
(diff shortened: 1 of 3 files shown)


lancerts:
@yitianlian

slime/utils/misc.py
```diff
@@ -65,3 +65,23 @@ def get_free_port(start_port=10000, consecutive=1):
     while not all(is_port_available(port + i) for i in range(consecutive)):
         port += 1
     return port
+
+
+def should_run_checkpoint_action(
```
yitianlian: The function name may change as it is not only used to check whether to run the checkpoint action, but also the evaluation action.

yitianlian (commented):
LGTM! Only need to change the func name~

slime/utils/misc.py
lancerts: sure, updated to `should_run_periodic_action` which covers (eval/save/checkpoint).

PR #1033, opened 2025-12-05
Improve debug output formatting in replay_reward_fn.py

lancerts:
Resolves TODO in replay_reward_fn.py by adding structured debug output with sample separators, field labels, and text truncation for better readability.


yitianlian (approved):
LGTM!

PR #1038, opened 2025-12-05
[Fix] Fix some bugs when on/offload model

yitianlian:


slime/backends/megatron_utils/actor.py
```diff
@@ -201,6 +202,16 @@ def _get_rollout_data(self, rollout_data_ref: Box) -> RolloutBatch:
             ]
         return rollout_data
 
+    @property
+    def active_model_tag(self) -> str | None:
+        return getattr(self, "_active_model_tag", None)
```
zhuzilin: this property doesn't seem necessary.

slime/backends/megatron_utils/actor.py
```diff
@@ -365,9 +374,9 @@ def train_actor(self, rollout_id: int, rollout_data: RolloutBatch) -> None:
                             os.environ["ROUTING_REPLAY_STAGE"] = "replay_forward"
                         else:
                             os.environ["ROUTING_REPLAY_STAGE"] = "record"
+                    self._switch_model("old_actor" if self.args.keep_old_actor else "actor")
```
zhuzilin: I think we can move this switch out of the `if not self.args.use_rollout_logprobs or self.args.get_mismatch_metrics`. And with such change, we no longer need to change the if below.

zhuzilin (approved):
LGTM! Left some minor comments.

PR #1042, opened 2025-12-06
[Draft] Update Megatron patch to work for Megatron v0.15.0

Birch-san:
Megatron 0.15.0 [came out yesterday](https://github.com/NVIDIA/Megatron-LM/releases/tag/core_v0.15.0).

I tried to rebase [megatron.patch](https://github.com/THUDM/slime/blob/b25cae6f42798627d0132d735284686cd4b8c7a8/docker/patch/latest/megatron.patch) onto it, but there were non-trivial merge conflicts. for example, Megatron now supports non-split qkv. so slime attention.py's "gated attention or not" condition needs some thought.

I took a guess at how to resolve the conflicts, here's what I came up with.  
the Megatron diff can be viewed directly here:
https://github.com/NVIDIA/Megatron-LM/compare/core_v0.15.0...NovelAI:Megatron-LM:v0.15.0-slime-patch

the `.patch` file here was computed by running this command inside that Megatron fork:

```bash
git diff core_v0.15.0 > changes.patch
```

**I haven't tried running it yet**, hence this is a draft. but perhaps this can save some time for anybody else who's looking into trying this update.

I wasn't sure why slime's Dockerfile convention is to apply a patch instead of checking out a branch of a forked repository. I wondered if maybe it was due to licensing considerations. so I tried to add a bit of licensing diligence as well (i.e. signpost slime's changes as Apache-2.0, and indicate files with derivative works in them). and include Pai-Megatron-Patch's NOTICE file (Apache 2.0 4.d. seemed to want NOTICEs to be distributed. it sounds like you can exclude portions that don't pertain to your derivative work, but it's easier to include the whole thing).  
I'm not sure if this is all that's required to satisfy the licenses, but it's probably directionally good.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,7 @@ FROM lmsysorg/sglang:${SGLANG_IMAGE_TAG} AS sglang
 # ======================================== Arguments =============================================
 
 ARG PATCH_VERSION=latest
-ARG MEGATRON_COMMIT=core_v0.14.0
+ARG MEGATRON_COMMIT=core_v0.15.0
 
 ARG ENABLE_CUDA_13=0
 


docker/patch/latest/megatron.patch
```diff
@@ -12,38 +61,16 @@ index 41c21d93d..ef80f72d6 100644
...
-index ccf5242a2..9b6d3e31f 100644
+index 5a1ea308d..87826e4df 100644
 --- a/megatron/core/dist_checkpointing/strategies/torch.py
 +++ b/megatron/core/dist_checkpointing/strategies/torch.py
-@@ -427,6 +427,15 @@ def _restore_dict_types(x: Union[dict, list, Any], keys_template: Union[dict, li
-             _restore_dict_types(x_val, templ_val)
- 
- 
-+@dataclass
-+class MCoreMetadata(Metadata):
```
Birch-san, line 23: for those surprised to see "why would we see any lines deleted when we rebase our patch onto a newer Megatron": I guess what happened here is that the MCoreMetadata dataclass got upstreamed? Megatron v0.15.0 already has this, so there's no longer a need for the patch to add it:

https://github.com/NVIDIA/Megatron-LM/blob/0112156741214c833a47d9f92911400ae96d724d/megatron/core/dist_checkpointing/strategies/torch.py#L462

zhuzilin:
Thank you for this and we are also considering upgrading megatron.

> I wasn't sure why slime's Dockerfile convention is to apply a patch instead of checking out a branch of a forked repository. I wondered if maybe it was due to licensing considerations

The reason is simply that we hope not to maintain a megatron fork by ourselves :p
And we don't need to add license to the patch at the moment (which can keep the patch a bit shorter).

> so slime attention.py's "gated attention or not" condition needs some thought.

I'm no sure if 0.15.0 have native gated attn support. If it does, we should use the native implementation. The origin one in the patch was a bit ad-hoc.

zhuzilin:
closed as we have already upgraded to megatron dev branch. But still thank you so much for the PR :heart:.

PR #1057, opened 2025-12-08
fix prepare_model_inputs for apply_chat_template=False

peterjc123:


peterjc123:
@yitianlian Could you please take a look?

Issue #1008, opened 2025-12-03
SFT for multi-turn tool call data is not supported properly


Surya-Gunukula:
# SFT Tool Support Proposal

## Problem

1. When `--apply-chat-template` is used, the message history (`list[dict]`) gets converted to a string in `data.py`. This breaks `gen_multi_turn_loss_mask_qwen` and other functions in `mask_utils.py`, which expect a `list[dict]`.

2. `--tool-key` requires `--apply-chat-template`, which triggers the same string conversion issue.

3. In `mask_utils.py`, `apply_chat_template` is called on each message individually without the `tools` parameter, so tool-related messages (assistant with `tool_calls`, `tool` role responses) are not formatted correctly.

## Downstream Effects

- Without these flags: Tool definitions are never added to the conversation, so the model never sees them during training.
- With these flags: A fatal error is thrown because `mask_utils` receives a string instead of `list[dict]`.
- Even if it didn't crash: `apply_chat_template` without `tools=` produces incorrect tokenization for tool-calling messages.

## Proposal

A non-invasive fix with two changes:

### 1. `data.py` - Store tools in metadata

When `--tool-key` is provided:
- Extract tools from the data
- Store them in `sample.metadata["tools"]`
- Keep messages as `list[dict]` (do NOT apply chat template)

### 2. `mask_utils.py` - Pass tools to `apply_chat_template`

Update the mask generation functions to:
- Accept an optional `tools` parameter
- Pass `tools` to `apply_chat_template` so tool-calling messages are formatted correctly

**Also this fixes Issue #289**

coding-famer:
I believe the change proposal 1. have been done by #501

Surya-Gunukula:
Will check out the new PR and make the appropriate fixes if there are any left. Thanks for bringing that up I wouldn't have noticed.

PR #1073, opened 2025-12-09
Add tool call support for multi-turn SFT with delta-based loss masking [FEATURE]

Surya-Gunukula:
### Summary

Adds support for training on multi-turn conversations with tool calls in SFT mode. This enables fine-tuning models on agentic datasets where assistant messages include `tool_calls` and conversations include `tool` role responses.

### Problem

The existing `MultiTurnLossMaskGenerator` uses a per-message tokenization approach that doesn't work with tool-call conversations. This is because `tools=...` is never passed to `apply_chat_template`, so the conversation never injects the tool call schema, so how is the model supposed to predict tool calls for SFT?

I also could not use the existing function as per-message tool messaging would inject the tool call schema into each message, so I opted for delta-based masking.

### Changes

**`slime/utils/mask_utils.py`**
- Added `gen_multi_turn_loss_mask_with_tools()` which gets called if `--tool-key` is applied
- This makes it non-harmful as SFT without `--tool-key` passed still works normally

**`slime/rollout/sft_rollout.py`**
- Extract `tools` from `sample.metadata` and pass to `get_loss_mask()`

### Usage

Dataset format (JSONL/Parquet):
```json
{
  "messages": [
    {"role": "user", "content": "What's the weather in NYC?"},
    {"role": "assistant", "content": null, "tool_calls": [{"id": "call_1", "type": "function", "function": {"name": "get_weather", "arguments": "{\"location\": \"NYC\"}"}}]},
    {"role": "tool", "tool_call_id": "call_1", "content": "Sunny, 72°F"},
    {"role": "assistant", "content": "The weather in NYC is sunny and 72°F!"}
  ],
  "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get weather", "parameters": {...}}}]
}
```

Run script args:
```bash
--input-key messages
--tool-key tools
```

### References

Delta-based tokenization approach adapted from:
- `examples/tau-bench/trainable_agents.py`
- [verl multi-turn documentation](https://verl.readthedocs.io/en/v0.4.1/sglang_multiturn/multiturn.html)

Also this finishes solving my issue #1008 (part 3, another PR already solved 1&2)

### DIFF
--- a/slime/utils/arguments.py
+++ b/slime/utils/arguments.py
@@ -1575,13 +1575,12 @@ def slime_validate_args(args):
         )
         args.eval_max_context_len = args.rollout_max_context_len
 
-    if args.rollout_max_prompt_len is None and args.rollout_max_context_len is not None:
-        logger.info(
-            f"args.rollout_max_prompt_len is not set. Use args.rollout_max_context_len - 1 ({args.rollout_max_context_len} - 1) as default value so that there is at least one generated token to compute loss."
-        )
-        args.rollout_max_prompt_len = args.rollout_max_context_len - 1
-
-    if args.rollout_max_prompt_len is not None and args.rollout_max_context_len is not None:
+    if args.rollout_max_context_len is not None:
+        if args.rollout_max_prompt_len is None:
+            args.rollout_max_prompt_len = args.rollout_max_context_len - 1
+            logger.info(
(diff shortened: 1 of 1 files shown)


Surya-Gunukula:
@zhuzilin could you please review this?

PR #1080, opened 2025-12-10
[Feature] PD Disaggregation Support

yitianlian:


zhuzilin:
nit: we don't need special logic to init sgl-router because sgl-router now supports init in pd disaggregation mode with empty prefill_urls and decode_urls.

PR #1086, opened 2025-12-11
Empty cache before finalize_model_grads to prevent unexpected oom

zhuzilin:
This is an ad-hoc fix for nccl failed to calloc error.

image

### DIFF
--- a/slime/backends/megatron_utils/model.py
+++ b/slime/backends/megatron_utils/model.py
@@ -500,6 +500,16 @@ def should_disable_forward_pre_hook(args: Namespace) -> bool:
     return args.use_distributed_optimizer and args.overlap_param_gather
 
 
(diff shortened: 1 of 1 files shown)


fzyzcjy:
brainstorms

1. do the check before every hijacked nccl call (since this is b/c torch reserve too much memory and when nccl wants some mem torch does not know and do not release anything for it, and b/c we re-create nccl in each iteration the badness is amplified)
2. --train-memory-margin-bytes
3. hack nccl mem allocation logic, when it finds no mem, tell torch to do a empty_cache

zhuzilin:
> --train-memory-margin-bytes

I didn't notice this flag, I will revert this change and set `--train-memory-margin-bytes` to 1 GB by default.

Fixed in #1088

PR #1113, opened 2025-12-15
fix: improve tokenization

nanjiangwill:
improve tokenization logics and update all examples


examples/search-r1/generate_with_search.py
Lins-01, line 1: > payload = {
            "text": prompt + response,
        }
        
       here, prompt is still of list type, maybe forgot to change it to prompt_text ?

nanjiangwill: good catch, updated

PR #1130, opened 2025-12-16
[Fix] Fix some bugs in retool example

yitianlian:
1. Change text-in token-out to token-in token-out
2. Add max context limitation.

### DIFF
--- a/examples/retool/generate_with_retool.py
+++ b/examples/retool/generate_with_retool.py
@@ -235,7 +235,7 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
         if args.rollout_max_context_len is not None:
             max_context_length = args.rollout_max_context_len
         else:
-            max_context_length = args.cp_size * args.max_tokens_per_gpu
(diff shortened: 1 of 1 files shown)


examples/retool/generate_with_retool.py
```diff
@@ -230,9 +230,20 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
...
     for turn in range(TOOL_CONFIGS["max_turns"]):
-        # Simple: just send prompt + response
+        # Check if total length exceeds max context length
+        total_length = len(prompt_tokens_ids) + len(response_token_ids)
+        if args.rollout_max_context_len is not None:
+            max_context_length = args.rollout_max_context_len
+        else:
+            max_context_length = args.context_parallel_size * args.max_tokens_per_gpu
+        if total_length >= max_context_length:
+            sample.status = Sample.Status.TRUNCATED
```
SwordFaith, line 240: Should we add a [:max_context_length] here?

PR #1134, opened 2025-12-16
Fix bug in dapo reward

ChangyiYang:


lilei199908:
> _ No description provided. _

you can set --reward-key to get reward

ChangyiYang:
Yes but that would make -rm-type dapo not out-of-box. If it's intended I can close this pr

PR #1135, opened 2025-12-17
[On Policy Distillation] resolve log prob dimension mismatch in on-policy distillation with CP > 1

Yuchen-Cao:
This PR addresses a dimension mismatch issue in on_policy_distillation when Context Parallel (CP) is enabled (cp_size > 1).

Problem:
Previously, the implementation did not support slicing the teacher's log prob according to the Context Parallel rank. While the student's log prob adhered to CP slicing, the teacher's log prob remained unsliced. This inconsistency caused a shape mismatch error between the student and teacher tensors during loss computation.

Solution:
I have updated the logic to apply CP slicing to the teacher's log prob, ensuring they align with the student's log prob. This fixes the dimension error and enables proper support for on-policy distillation with Context Parallelism.

### DIFF
--- a/slime/backends/megatron_utils/actor.py
+++ b/slime/backends/megatron_utils/actor.py
@@ -215,7 +215,10 @@ class MegatronTrainRayActor(TrainRayActor):
                     dtype=torch.float32,
                 )
                 for log_prob, total_length, response_length in zip(
-                    rollout_data["teacher_log_probs"], rollout_data["total_lengths"], rollout_data["response_lengths"]
+                    rollout_data["teacher_log_probs"], 
+                    rollout_data["total_lengths"], 
+                    rollout_data["response_lengths"],
+                    strict=False,
                 )
             ]
         if "rollout_routed_experts" in rollout_data:


slime/backends/megatron_utils/loss.py
```diff
@@ -288,14 +288,12 @@ def compute_advantages_and_returns(args: Namespace, rollout_data: RolloutBatch)
...
-        teacher_log_probs = [t_log_prob.to(device=device) for t_log_prob in teacher_log_probs]
-        teacher_log_probs = [
-            t_log_prob[-response_length:]
-            for t_log_prob, response_length in zip(teacher_log_probs, response_lengths, strict=False)
-        ]
+        teacher_log_probs = [t.to(device=device) for t in teacher_log_probs]
         advantages = [
-            teacher_log_prob - student_log_prob
-            for teacher_log_prob, student_log_prob in zip(teacher_log_probs, student_log_probs, strict=False)
+            t_log_prob[-response_length:] - s_log_prob[-response_length:]
```
yitianlian: I think we shouldn't slice logp with `response_length` because the current logp is CP sliced?

Yuchen-Cao: Oh yes!

Just some previous attempts when i tried to match the size between them. will change this line back.

Yuchen-Cao: Done

yitianlian:
Maybe you should run the pre-commit to fix the format error.

PR #1140, opened 2025-12-17
[FSDP][1/n] Support LoRA training for FSDP backend.

GuanxingLu:
Approaches Considered:

1. Merge/Unmerge: Merge LoRA to base -> send merged weights -> unmerge for training.
2. File-based (Current): Save adapter to disk -> rollout engine loads via load_lora_adapter (e.g., AReal).
3. Direct Tensor: Send tensors via memory (e.g., verl).

Current Implementation: Implemented Option 2 using UpdateWeightFromDisk.

Limitation: When rollout engine offloading is enabled, the base model should be re-transmitted in each update (though overhead is manageable).

Status:
✅ Train Qwen3-4B with LoRA on single A100 40G (Full param triggers OOM).
🚧 Reward curve pending.

TODO:
- [ ] Implement Options 1 & 3 for comparison.

- [ ] Save/Load logic. 
- [ ] Efficient layer summon for LoRA weight update. 
- [ ] Option 3: Send LoRA tensors rather than files.

Discussion Feedback on the trade-offs of these approaches is welcome!

### DIFF
--- a/slime/backends/fsdp_utils/actor.py
+++ b/slime/backends/fsdp_utils/actor.py
@@ -141,10 +141,11 @@ class FSDPTrainRayActor(TrainRayActor):
         if with_ref:
             self.ref_model = self._create_ref_model(args.ref_load)
 
-        if self.args.colocate:
-            self.weight_updater = UpdateWeightFromTensor(self.args, self.model)
-        else:
-            self.weight_updater = UpdateWeightFromDistributed(self.args, self.model)
+        self.weight_updater = (
+            UpdateWeightFromTensor(self.args, self.model)
+            if self.args.colocate
+            else UpdateWeightFromDistributed(self.args, self.model)
+        )
 
         checkpoint.finalize_load(self, checkpoint_payload)
 
--- a/slime/backends/fsdp_utils/actor.py
+++ b/slime/backends/fsdp_utils/actor.py
@@ -29,6 +29,7 @@ from ...utils import tracking_utils
 from ...utils.profile_utils import TrainProfiler
 from . import checkpoint
 from .data_packing import pack_sequences, pad_packed_sequence_with_cp, unpack_sequences
+from .lora_utils import apply_lora_to_model, is_lora_model
 from .lr_scheduler import get_lr_scheduler
 from .update_weight_utils import UpdateWeightFromDistributed, UpdateWeightFromTensor
 
@@ -95,6 +96,9 @@ class FSDPTrainRayActor(TrainRayActor):
                 attn_implementation=self.args.attn_implementation,
             )
 
+        if self.args.lora_rank > 0 or self.args.lora_adapter_path:
+            model = apply_lora_to_model(model, self.args)
+
         model.train()
 
         full_state = model.state_dict()
@@ -108,11 +112,14 @@ class FSDPTrainRayActor(TrainRayActor):
         self.model = model
 
         if args.gradient_checkpointing:
-            self.model.gradient_checkpointing_enable()
+            # Avoid "does not require grad" error
+            gc_kwargs = {"use_reentrant": False} if is_lora_model(self.model) else {}
+            self.model.gradient_checkpointing_enable(gradient_checkpointing_kwargs=gc_kwargs)
 
         if args.optimizer == "adam":
+            trainable_params = [p for p in self.model.parameters() if p.requires_grad]
             self.optimizer = torch.optim.AdamW(
-                self.model.parameters(),
+                trainable_params,
                 lr=args.lr,
                 betas=(args.adam_beta1, args.adam_beta2),
                 eps=args.adam_eps,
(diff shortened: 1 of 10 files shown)


PopSoda2002:
Great work!

slime/backends/sglang_utils/sglang_engine.py
```diff
@@ -10,6 +10,7 @@
 from sglang.srt.utils import kill_process_tree
 from urllib3.exceptions import NewConnectionError
 
+from slime.backends.fsdp_utils.lora_utils import get_explicit_target_modules
```
PopSoda2002: I suggest make the `get_explicit_target_modules` to somewhere else, it is weird for sglang depends on a utils func in fsdp

slime/backends/fsdp_utils/update_weight_utils.py
```diff
@@ -256,3 +264,84 @@ def update_bucket_weights(self, named_tensors, weight_version=None) -> None:
...
+
+        save_lora_to_disk(self.model, self._lora_save_dir)
+
+        dist.barrier()
+
+        if dist.get_rank() == 0:
+
+            wait_for_lora_ready(self._lora_save_dir)
+
+            if self._lora_loaded:
```
PopSoda2002: I think we can merge `_base_sync_done` and `_lora_loaded` into one variable?

PopSoda2002 (commented):
Has some nit for you :)

slime/backends/fsdp_utils/update_weight_utils.py
GuanxingLu: Sure. I will merge them.

slime/backends/sglang_utils/sglang_engine.py
GuanxingLu: Yeh, I can prepare the target_modules in advance to avoid processing them in sglang_engine.py

GuanxingLu:
@PopSoda2002 Could you please check this? Thank you!

Current Status:
- Add '--offload-rollout-level' argument, level 1: offload kv cache/cuda graph only, level 2: offload weights + kv cache/cuda graph. When the rollout engine  does not release the base model weights, we only need to sync LoRA weights.
- I modified `UpdateWeight` base class (this should be invasive) to support LoRA update alongside base model weight update.
- Current LoRA weight update logic works for both colocated and disaggregated modes. 
- Fixed nits you mentioned before. But I can not remove _lora_loaded because unload_lora_adapter will throw error if LoRA is not loaded.

More TODOs:
[] Save/Load logic.
[] Efficient layer summon for LoRA weight update.
[] Option 3: Send LoRA tensors rather than files.

PopSoda2002:
> @PopSoda2002 Could you please check this? Thank you!
> 
> Current Status:
> 
> * Add '--offload-rollout-level' argument, level 1: offload kv cache/cuda graph only, level 2: offload weights + kv cache/cuda graph. When the rollout engine  does not release the base model weights, we only need to sync LoRA weights.
> * I modified `UpdateWeight` base class (this should be invasive) to support LoRA update alongside base model weight update.
> * Current LoRA weight update logic works for both colocated and disaggregated modes.
> * Fixed nits you mentioned before. But I can not remove _lora_loaded because unload_lora_adapter will throw error if LoRA is not loaded.
> 
> More TODOs: [] Save/Load logic. [] Efficient layer summon for LoRA weight update. [] Option 3: Send LoRA tensors rather than files.

I moved these to the PR description

PopSoda2002:
Can you have some test case in wandb compared with no lora? I can help with this part

GuanxingLu:
Sure, you can run the test. I am also running them in my local setting.

PopSoda2002:
@GuanxingLu And also, can you please fix the linting? You can use `pre-commit`

GuanxingLu:
Sorry, now `pre-commit` tests are passed.

slime/backends/fsdp_utils/actor.py
```diff
@@ -134,11 +141,10 @@ def init(self, args: Namespace, role: str, with_ref: bool = False) -> int:  # ty
...
 
-        self.weight_updater = (
-            UpdateWeightFromTensor(self.args, self.model)
-            if self.args.colocate
-            else UpdateWeightFromDistributed(self.args, self.model)
-        )
+        if self.args.colocate:
+            self.weight_updater = UpdateWeightFromTensor(self.args, self.model)
+        else:
+            self.weight_updater = UpdateWeightFromDistributed(self.args, self.model)
```
PopSoda2002: Can I ask why change this?

slime/backends/fsdp_utils/update_weight_utils.py
```diff
@@ -45,31 +48,77 @@ def connect_rollout_engines(
...
+
+        if dist.get_rank() == 0:
+            if self._lora_loaded:
+                refs = [engine.unload_lora_adapter.remote(LORA_ADAPTER_NAME) for engine in self.rollout_engines]
+                ray.get(refs)
+
+            refs = [engine.flush_cache.remote() for engine in self.rollout_engines]
+            ray.get(refs)
+
+            refs = [
```
Hecate0821, line 110: I am curious how long it take for SGLang to read LoRA weights from disk. Is it possible to pass through NCCL？I am not sure about the file size

Hecate0821: And we can not assume that in distributed training, there is a shared file system for every node. So the read from disk approach may not work here

tellyoung:
It seems that some examples are still missing

arminZolfaghari:
Hi @GuanxingLu, thanks for working on this. 

I'm planning to run LoRA RL on Qwen3-8B and 14B with the `examples/retool` workflow on 4×GH200 (full-param OOMs on this box). Two questions:

1. Has anyone tried this branch with `examples/retool`, or only the math GRPO scripts?
2. Of the remaining TODOs (save/load, efficient layer summon, tensor-based sync), which are planned for `[2/n]` vs later PRs?

Happy to test the branch on 4×GH200 and report back if that's useful. 
Thanks!

Issue #1075, opened 2025-12-09
[Roadmap][VLM] Support VLM Multi-Turn


nanjiangwill:
# TODOs for VLM Multi-Turn Support
Generalize multi-turn support for both **FSDP** and **Megatron**.  
All implementation should remain under `examples/`.

Single-turn reference: https://github.com/THUDM/slime/pull/501

## Dataset
- [ ] Add support for **OpenCUA**: https://opencua.xlang.ai/

## Rollout (`--custom-generate-function-path`)
Create a custom rollout function to support multi-turn interactions.

### Multi-Turn Config
- [ ] Add configs such as `max_turns`.
- [ ] Add early-stop logic (e.g., stop on max tokens).

### Loss Mask
(ref: https://github.com/THUDM/slime/blob/v0.2.0/slime/ray/rollout.py#L236).

Potential behavior:
- Prepare `tokens` and `loss_mask` **inside the custom rollout**.
- Initialize with empty `tokens` and empty `loss_mask`.
- For each **input** turn:
  - Append input tokens to `tokens`.
  - Append `0`s of equal length to `loss_mask`.
- For each **generated** output:
  - Append output tokens to `tokens`.
  - Append `1`s of equal length to `loss_mask`.

gxlvera:
Hi, I want to take this

simondong1:
Hi, I would love to contribute

jhinpan:
Feel free to contribute! If you run into any issues you’d like to discuss, just leave a comment here. Once you have a PR or draft ready, you can ping me and @nanjiangwill for review.

HaZ-K:
Hi, I want to  contribute too

kafkayu:
Hi, I would like to contribute

Liac-li:
work work

gxlvera:
> Hi, I want to take this



Here is my draft PR: https://github.com/THUDM/slime/pull/1141

PR #1141, opened 2025-12-17
[VLM] end2end geo3k multi-turn RL of VLM Recipe

gxlvera:
## Goal
VLM Multi-turn (related to https://github.com/THUDM/slime/issues/1075)

### Rollout
- [x] Created a custom multi-turn rollout function in `examples/vlm_multi_turn/rollout.py`
  - [x] Pluggable interactive env (env path specified via rollout argument --custom-config-path)
  - [x] Early-stop logic
    - [x] `max_turns` (specified via rollout argument --custom-config-path)
    - [x] max_new_token cap
  - [x] `loss_mask` / `rollout_log_probs`
    - `loss_mask = 1` on assistant tokens
    - `loss_mask = 0` on user/observation tokens
    - `rollout_log_probs` padded to match
    - initial `sample.prompt` stays unmasked

### Interactive environment
- [x] Custom env split from rollout: `examples/vlm_multi_turn/env_geo3k.py`
  - `build_env`/ `reset` / `step` / `format_observation` functions for per-turn feedback

### Data & dataset
- [x] Support geo3k multimodal dataset: https://huggingface.co/datasets/chenhegu/geo3k_imgurl
- [ ] OpenCUA not supported yet (may require a more complex interactive env, e.g. https://github.com/xlang-ai/OSWorld)

## Experiment Result
Trained Qwen3-VL-2B-Instruct on the geo3k dataset with multi-turn reasoning, using GRPO. 
final_result

### DIFF
--- a/examples/geo3k_vlm_multi_turn/rollout.py
+++ b/examples/geo3k_vlm_multi_turn/rollout.py
@@ -7,13 +7,13 @@ from pathlib import Path
 from typing import Any
 
 import torch
+from examples.geo3k_vlm_multi_turn.base_env import BaseInteractionEnv
 
 # When executed as a module: python -m examples.vlm_multi_turn.rollout
 from slime.rollout.sglang_rollout import GenerateState
 from slime.utils.http_utils import post
 from slime.utils.processing_utils import encode_image_for_rollout_engine
 from slime.utils.types import Sample
-
 
 DEFAULT_ENV_MODULE = "examples.vlm_multi_turn.env_geo3k"
 
@@ -207,7 +207,7 @@ async def _run_inference_step(url: str, tokens: list[int], sampling_params: dict
     return response_text, new_tokens, new_log_probs, finish_type
 
 
-def _process_env_step(env, response_text: str, tokenizer, processor, args, sample_metadata):
+def _process_env_step(env: BaseInteractionEnv, response_text: str, tokenizer, processor, args, sample_metadata):
     observation, done, _ = env.step(response_text)
     if done:
         return None, None, None, None, True


slime/utils/arguments.py
```diff
@@ -199,6 +199,13 @@ def add_rollout_arguments(parser):
                     "and `truncated`."
                 ),
             )
+            parser.add_argument(
+                "--max-turns",
+                dest="max_turns",
+                type=int,
+                default=None,
+                help="Maximum turns for multi-turn custom rollout (e.g., Sokoban). Defaults to rollout implementation config.",
+            )
```
zhuzilin: is it possible to pass these 2 configs through `--custom-config-path`?

gxlvera: Yes, it sounds neater. I have pushed the change, thanks!

zhaochenyang20:
Nice done Xiaole!

ygxio:
@gxlvera are you working on OpenCUA part? can i help with it?

zhaochenyang20:
Great job so far!

gxlvera:
> @gxlvera are you working on OpenCUA part? can i help with it?

Hi, you could try to support OpenCUA's AgentNet [dataset](https://huggingface.co/datasets/xlangai/AgentNet). Note that if you want to implement the online interaction, maybe you need an os sandbox for simulation. It's OK if you stick with offline mode (without interaction) although I personally don't think that would work well.

ygxio:
Sure @gxlvera can try that

ygxio:
@gxlvera can you help with openCUA a bit?  where can i connect u?

gxlvera:
> @gxlvera can you help with openCUA a bit? where can i connect u?

Hi, you could DM me at gxlvera@gmail.com~

examples/geo3k_vlm_multi_turn/env_geo3k.py
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        return obs, is_final_turn, info
+
+
+def _extract_ground_truth(sample: Sample | None) -> str | None:
+    """Resolve the ground-truth answer from label or metadata."""
+    if sample is None:
+        return None
+    if getattr(sample, "label", None) is not None:
```
zhaochenyang20: Do not use `getattr` as much as possible. Change it to:

`if sample.label is not None:`

zhaochenyang20 (changes requested):

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -0,0 +1,321 @@
...
+            continue
+        if isinstance(merged[key], torch.Tensor) and isinstance(val, torch.Tensor):
+            merged[key] = torch.cat([merged[key], val], dim=0)
+        elif isinstance(merged[key], list) and isinstance(val, list):
+            merged[key] = merged[key] + val
+        elif isinstance(merged[key], dict) and isinstance(val, dict):
+            merged[key] = {**merged[key], **val}
+        else:
+            merged[key] = val
+    return merged
```
zhaochenyang20: Do not perform torch.cat inside the rollout loop. As the conversation grows, this results in $O(N^2)$ memory allocation and copying overhead. This causes significant latency spikes and GPU memory fragmentation.

Please store the tensors in a list during the loop. Perform a single torch.cat or stack only once when constructing the final Sample object.

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -0,0 +1,321 @@
...
+            obs_prompt_ids, obs_image_data, obs_multimodal_inputs, obs_multimodal_train_inputs = (
+                _encode_observation_for_generation(
+                    tokenizer,
+                    processor,
+                    next_user_message,
+                    sample.metadata,
+                    getattr(args, "apply_chat_template", False),
+                    args.apply_chat_template_kwargs,
+                )
+            )
```
zhaochenyang20: If possible, cache the processed vision features of the static images and only process the new text/tools delta?

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -0,0 +1,321 @@
...
+
+            # Drop a leading BOS if present to avoid injecting it mid-stream.
+            bos_id = getattr(tokenizer, "bos_token_id", None)
```
zhaochenyang20: The code relies heavily on getattr(args, ...) and getattr(sample, ...). This defeats static type checking and makes the code fragile (hard to refactor).

examples/vlm_multi_turn/rollout.py
```diff
@@ -0,0 +1,321 @@
...
+# When executed as a module: python -m examples.vlm_multi_turn.rollout
+from slime.rollout.sglang_rollout import GenerateState
+from slime.utils.http_utils import post
+from slime.utils.processing_utils import encode_image_for_rollout_engine
+from slime.utils.types import Sample
+
+
+DEFAULT_ENV_MODULE = "examples.vlm_multi_turn.env_geo3k"
+DEFAULT_ROLLOUT_CONFIG = {
+    "max_turns": 5,
```
zhaochenyang20: DEFAULT_ROLLOUT_CONFIG has max_turns=5 in rollout.py but max_turns=3 in env_geo3k.py.

please unify them.

examples/geo3k_vlm_multi_turn/env_geo3k.py
```diff
@@ -0,0 +1,293 @@
...
+        if boxed:
+            return str(boxed).strip()
+        for line in reversed(text.splitlines()):
+            cleaned = line.strip()
+            if cleaned:
+                return cleaned[:512]
+        trimmed = text.strip()
+        return trimmed[:512] if trimmed else None
+
+    def _extract_balanced_json(self, text: str, start: int) -> str | None:
```
zhaochenyang20, line 125: Instead of parsing raw strings manually, use a robust library like `dirtyjson` or enforce a strict regex extraction for the <tool_call> block contents. The model should be trained to output valid JSON.

examples/geo3k_vlm_multi_turn/rollout.py
zhaochenyang20: maybe do it like:

```python
multimodal_inputs_buffer = []

if sample.multimodal_train_inputs:
    multimodal_inputs_buffer.append(sample.multimodal_train_inputs)

for turn_idx in range(max_turns):
    
    if obs_multimodal_train_inputs:
        # only Append, do not Merge
        multimodal_inputs_buffer.append(obs_multimodal_train_inputs)
```

examples/geo3k_vlm_multi_turn/rollout.py
```diff
@@ -0,0 +1,321 @@
...
+        elif isinstance(merged[key], list) and isinstance(val, list):
+            merged[key] = merged[key] + val
+        elif isinstance(merged[key], dict) and isinstance(val, dict):
+            merged[key] = {**merged[key], **val}
+        else:
+            merged[key] = val
+    return merged
+
+
+async def generate(args: Any, sample: Sample, sampling_params) -> Sample:
```
zhaochenyang20: This function is too huge to review and maintain. Please split it into several parts, like:

1. _prepare_initial_inputs(sample, processor, tokenizer)

 Handles initial encoding.

2. _run_inference_step(url, tokens, sampling_params, image_data, tokenizer)

Handles model inference IO. 

3. _process_env_step(env, env_module, response_text, tokenizer, processor, args, sample_metadata)

Handles environment interaction & observation encoding.

4. _update_multimodal_state(sample, current_image_data, obs_imgs, obs_mm, obs_mm_train)

Handles multimodal state synchronization.

and then, the function is like:

```python
async def generate(args: Any, sample: Sample, params: dict) -> Sample:
    """Minimalist multi-turn rollout orchestration."""
    # 1. Initialization
    env, config, state, url = _initialize_resources(args, sample)
    img_data, budget = _prepare_start_state(sample, state, args, params)

    try:
        env.reset()
        for _ in range(config["max_turns"]):
            
            # --- A. Model Inference (Assistant Turn) ---
            txt, ids, probs, finish = await _run_inference_step(
                url, sample.tokens, params, img_data, state.tokenizer
            )
            
            # Update State: Mask=1 for model outputs (Compute Loss)
            _append_to_sample(sample, ids, probs, loss_mask_val=1)
            
            if _should_stop(sample, finish, budget, len(ids)): break

            # --- B. Environment Interaction (User/System Turn) ---
            obs_ids, obs_imgs, obs_mm, obs_train, done = _process_env_step(
                env, config, txt, state.tokenizer, state.processor, args, sample.metadata
            )
            
            if done: 
                sample.status = Sample.Status.COMPLETED; break

            # Update State: Mask=0 for environment outputs (Ignore Loss)
            _append_to_sample(sample, obs_ids, [0.0]*len(obs_ids), loss_mask_val=0)
            
            # --- C. Sync Multimodal Data ---
            img_data = _update_multimodal_state(sample, img_data, obs_imgs, obs_mm, obs_train)

            if _should_stop(sample, None, budget, len(obs_ids)): break

        return _finalize_sample(sample, state.tokenizer)
        
    finally:
        env.close()
```

examples/vlm_multi_turn/rollout.py
gxlvera: Thanks for pointing out. I have unified them. But actually, I personally don't think it's necessary to do so. The logic of setting max_turn is 
1. check user's argument passed as `--custom-config-path`, if None
2. check env's max_turn, if None
3. fall back to the default max_turn set in rollout.py

So it's up to user's input and the specific env's setting.

examples/geo3k_vlm_multi_turn/rollout.py
gxlvera: What this function does is exactly only process the new content. It's just that the delta content might also include new multimodal data (e.g. for dataset like OpenCUA's GUI-agent trajectories, the env's response is usually a screenshot of the new computer UI). I have cached the multimodal data in previous turns, but we still need to process the new multimodal data (if any) in later turns.

examples/vlm_multi_turn/rollout.py
gxlvera: I further discussed with Nan Jiang and decided to delete the default rollout/env config. It's better to force user to input configuration parameter like max_turns for clearer logging and debugging

examples/geo3k_vlm_multi_turn/env_geo3k.py
gxlvera: Thanks, i have fixed it!

examples/geo3k_vlm_multi_turn/rollout.py
gxlvera: I have minimized the usage of it, only using it when necessary. In this PR, I now only use getattr() in rollout.py 

```python
def _format_observation(env_module, observation: dict) -> dict:
"""Convert an environment observation into a chat message."""
    formatter = getattr(env_module, "format_observation", None)
    if callable(formatter):
        return formatter(observation)
```

Since it's optional for the environment to provide the formatter, I think it's better to use getattr() here.

examples/geo3k_vlm_multi_turn/rollout.py
gxlvera: Thanks for the valuable advice! I have implemented the optimized way as below
```python
def _merge_multimodal_train_inputs(chunks: list[dict | None]) -> dict | None:
    """
    Merge per-turn multimodal_train_inputs with a single concat per key.
    """
    if not chunks:
        return None

    values_by_key = {}
    for chunk in chunks:
        if not chunk:
            continue
        for key, val in chunk.items():
            if val is None:
                continue
            values_by_key.setdefault(key, []).append(val)

    merged = {}
    for key, values in values_by_key.items():
        if all(isinstance(v, torch.Tensor) for v in values):
            merged[key] = torch.cat(values, dim=0)

    return merged

for turn_idx in range(max_turns):
    if obs_multimodal_train_inputs:
        multimodal_train_inputs_buffer.append(obs_multimodal_train_inputs)
```

examples/geo3k_vlm_multi_turn/env_geo3k.py
gxlvera: Make sense, i have implemented the robust json extraction as below. (I used orjson instead of dirtyjson is because dirtyjson is not installed by slime's environment) :
```python
def _parse_tool_payload(self, raw_json: str) -> dict[str, Any] | None:
        """Parse tool payload strictly as JSON. Malformed payloads are rejected."""
        loader = orjson.loads
        try:
            return loader(raw_json)
        except Exception as exc:
            logger.warning("Failed to decode tool call payload: %s", exc)
            return None

def _extract_tool_call(self, text: str) -> dict[str, Any] | None:
    #some unmodified codes are skipped
    payload = self._parse_tool_payload(raw_json)
        if payload is None:
            return None
    name = payload.get("name") or payload.get("function", {}).get("name")
    arguments = payload.get("arguments") or payload.get("function", {}).get("arguments") or {}
    if isinstance(arguments, str):
        try:
            arguments = json.loads(arguments)
        except json.JSONDecodeError:
            logger.warning("Tool call arguments are not valid JSON; rejecting tool call.")
            return None
```

examples/geo3k_vlm_multi_turn/rollout.py
gxlvera: Indeed...😭 I have refactored the [generate function](https://github.com/THUDM/slime/pull/1141/files#diff-120dbcbc06d89becc26b8d53312f6ea7099f5fb015450a6631edaa14c9c94e14)

examples/geo3k_vlm_multi_turn/rollout.py
zhaochenyang20: getattr can and should be replaced.

zhaochenyang20: I eliminated the brittle reliance on getattr and runtime reflection by establishing an explicit OOP contract:

1. Define a BaseInteractionEnv class containing the default format_observation logic and ensure Geo3kEnv inherits from it, which guarantees that the method is always available on the instance and allows me to call env.format_observation() directly.

examples/geo3k_vlm_multi_turn/rollout.py
zhaochenyang20: Great job

zhaochenyang20:
We shall also have a Megatron version,. But FSDP works cool!

zhaochenyang20:
I evaluated the performance, which works well to me.

image

Issue #1157, opened 2025-12-19
[tau-bench] Make example runnable offline: stub user provider + robust tool parsing (parse_tools signature compatibility)


Fengzdadi:
## What happened

I’m running `examples/tau-bench/generate_with_tau.py` to validate the episode logging pipeline on a local machine (no GPU / no triton).
The run previously aborted due to:

- user simulation calling external LLM (litellm/gemini) → 429 / API key issues
- tool parser API mismatch (`parse_tools()` signature differences across versions)
- a small bug in response dict access (`output.get[...]` vs `output.get(...)`)

## Expected

- Be able to run **1 task end-to-end without external API keys** (at least for validating logging + control flow).
- Robust tool parsing fallback if tool parser is unavailable or signature differs.

## Proposed changes

1) **Offline user sim**
- Allow `user_model_provider="stub"` (or `provider="stub"`) so env step doesn’t call external LLM
- Returns a canned user message like `"(stub user) OK."`
- Cost = 0

2) **Tool parsing compatibility**
Add a wrapper in `openai_tool_adapter.py` that:
- detects `parse_tools` signature (2-arg vs 3-arg)
- supports different return schemas (e.g. missing `normal_text`)
- falls back to “no tool calls” when parser unavailable

3) **Bugfix**
Use:
```py
raw_response = output.get("text", "")
```
(currently it is `output.get["text"]`)

## Repro
- Start local sglang http endpoint (or stub server) on 127.0.0.1:30000
- Environment: macOS, CPU-only, no triton installed
- Sanity check:
``` bash
curl -s -X POST http://127.0.0.1:30000/generate \
  -H "Content-Type: application/json" \
  -d '{"text":"Say ONLY: OK123","sampling_params":{"temperature":0,"max_new_tokens":8}}'
```
- Run the tau-bench example with user_model_provider="stub"

## Result after patch
- episode.jsonl is generated and env_step proceeds with observation (stub user) OK.
- No external API calls are needed (no Gemini/OpenAI keys), so no 429.

PR #1158, opened 2025-12-19
tau-bench: offline stub user + tool parsing fallback

Fengzdadi:
## Summary
This PR makes the `examples/tau-bench` pipeline runnable end-to-end on CPU-only machines / CI without external LLM API keys, mainly for validating episode logging + control flow.

## Motivation
When running `examples/tau-bench/generate_with_tau.py` locally (no GPU / no triton), the run could abort due to:
- User simulation calling external LLM (e.g. LiteLLM/Gemini) → 429 / API key requirements
- `parse_tools()` API differences across versions (signature / return schema)
- Minor bug in dict access (`output.get[...]` vs `output.get(...)`)

## Changes
1) **Offline stub user simulation**
- Support `user_model_provider="stub"` to avoid external LLM calls during env reset/step
- Returns a canned message like `"(stub user) OK."`, cost=0

2) **Tool parsing compatibility**
- Add a compatibility wrapper for tool parsing to handle:
  - `parse_tools` signature differences (2-arg vs 3-arg)
  - missing fields like `normal_text`
  - fallback to “no tool calls” when tool parser is unavailable

3) **Logging + small bugfixes**
- Fix raw response access (`output.get("text", "")`)
- Improve episode logging robustness

## Testing
- `pre-commit run --all-files`
- Local run (CPU-only):
  - start local sglang http endpoint on `127.0.0.1:30000`
  - run with `user_model_provider="stub"`
  - confirmed `episode.jsonl` is generated and env_step proceeds with observation `(stub user) OK.`

## Notes
- This is intended for local debugging / CI sanity checks; it does not aim to replace full GPU-based evaluation.

## Related
- Fixes/addresses #1157

### DIFF
--- a/examples/tau-bench/episode_logger.py
+++ b/examples/tau-bench/episode_logger.py
@@ -1,0 +1,61 @@
+# examples/tau-bench/episode_logger.py
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+import time
+from dataclasses import dataclass
+from typing import Any
+
+
+def _truncate(s: str | None, max_chars: int = 8000) -> str | None:
+    if s is None:
+        return None
+    if len(s) <= max_chars:
+        return s
+    return s[:max_chars] + f"\n...[truncated {len(s)-max_chars} chars]"
+
+
+def _sha256_text(s: str) -> str:
+    return hashlib.sha256(s.encode("utf-8", errors="ignore")).hexdigest()
+
+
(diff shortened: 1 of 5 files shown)


zhuzilin:
Close as having  a fully functioned tau-bench test is not the main purpose of the examples in slime, which should be used to show some specific features.

PR #1170, opened 2025-12-21
[Bug] Fix raw_reward==0 when getting deepscaler

Zhuohao-Li:
result (blue: before; red: after):
image


zhuzilin:
please use --rm-type math if format loss is not needed.

Zhuohao-Li:
> please use --rm-type math if format loss is not needed.

make sense, close this PR

Issue #1144, opened 2025-12-18
[FSDP] Add fused kernel for MoE


Hecate0821:
Current FSDP backend MoE training is extremely slow. We could potentially utilize [sonic-moe](https://github.com/Dao-AILab/sonic-moe) to optimize it.

ChangyiYang:
Let's gooooooooooooooooooooooooooooo! Take it together with @Hecate0821

PR #1176, opened 2025-12-22
Integrate Sonic-Moe in FSDP

ChangyiYang:
As titled
issue https://github.com/THUDM/slime/issues/1144

### DIFF
--- a/moe-fresh.sh
+++ b/moe-fresh.sh
@@ -1,0 +1,158 @@
+#!/bin/bash
+
+# for rerun the task
+pkill -9 sglang
+sleep 3
+ray stop --force
+pkill -9 ray
+pkill -9 python
+sleep 3
+pkill -9 ray
+pkill -9 python
+pkill -9 redis
+
+set -ex
+
+# will prevent ray from buffering stdout/stderr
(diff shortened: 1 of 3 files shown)


junjzhang:
Hi, happended to see this pr, I guess the numerical accuracy is not correct. Sonic-moe will require a interleaved weight like below
```python
 I = w1.shape[1] // 2
# Interleave gate/up: [E, 2*I, H] concat -> [E, 2*I, H] interleaved
w1 = torch.stack([w1[:, :I, :], w1[:, I:, :]], dim=2).reshape(w1.shape[0], 2 * I, w1.shape[2])
```
We tested this in our own training framework, I suppose this is the correct way. BTW, we do not inspect significant improvement in MFU. Happy to discuss your results.

ChangyiYang:
Thanks for the detailed feedback! This makes sense — I’ve indeed observed some nonsensical outputs and am currently debugging the numerical correctness. I’ll try the interleaved gate/up weight layout you suggested and will update the PR description with performance results once ready.

faresobeid:
Any updates on this?

PR #1193, opened 2025-12-23
update quick start doc

zijiexia:
As per discussion with @zhaochenyang20 , update the quick start. Some changes has already been made on the slime side, so the remaining changes are following:
1. Since we already installed miles when building docker images, change the document to pull the latest code instead of clone the repo
2. Remove pip install -U huggingface_hub in the instructions since pip's dependency resolver does not take into account all the packages that are installed and will install a huggingface_hub version that is incompatible with other dependencies.


zhaochenyang20:
let me see this in 24h

PR #1200, opened 2025-12-24
fix: replace blocking sleep with async sleep and fix file handle leak

lancerts:
## Summary
- Async sleep fix (radix_tree_middleware.py): Replaced blocking time.sleep(30) with await asyncio.sleep(30) in the async dispatch method to prevent blocking the event loop.
- File handle leak fix (ray/rollout.py): Removed unnecessary open() call and pass path directly to torch.load(), which handles file I/O internally.


lancerts:
@yitianlian ty

PR #1208, opened 2025-12-25
Add ci for pd disaggregation

zhuzilin:


zhuzilin:
close as it seems that there is issue with running mooncake + rdma in the test nodes.

PR #1232, opened 2025-12-26
Revert data processing of VLM

zhuzilin:
fixing #https://github.com/THUDM/slime/pull/501#issuecomment-3692816459

Also, we are adding a type check in `generate` to make sure the other changes won't break custom generation

### DIFF
--- a/slime/utils/data.py
+++ b/slime/utils/data.py
@@ -164,6 +164,7 @@ class Dataset:
                 # temporary solution, will write image utils for slime later
                 from qwen_vl_utils import process_vision_info
 
+                assert isinstance(prompt, list)
                 images, videos = process_vision_info(prompt)
                 multimodal_inputs = {"images": images, "videos": videos}
             else:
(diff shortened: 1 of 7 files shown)


slime/utils/data.py
```diff
@@ -149,17 +149,36 @@ def __init__(
...
+                    **(apply_chat_template_kwargs or {}),
+                )
+            else:
+                formatted_prompt = prompt
+
+            if processor:
+                # temporary solution, will write image utils for slime later
+                from qwen_vl_utils import process_vision_info
+
+                images, videos = process_vision_info(prompt)
```
nanjiangwill, line 168: `process_vision_info` only takes list, so probably we need to add an assertion here

zhuzilin: fixed.

GentleZhu:
Curious if this breaks the VLM sft_rollout? Didn't fully understand why the VLM data processing is reverted?

GentleZhu:
In my experiment, this revert causes SFT on VLM failing to load image tokens

PR #1234, opened 2025-12-26
[VLM] optimize VLM processing

nanjiangwill:
followup #1232 

1. Remove extra processing of multimodal train inputs
2. Add CI for VLM FSDP
3. Update `--rotary-base` for VL model


zhuzilin:
I think that it's a.bit hard to understand the meaning of multimodal tensor, could you rename the multimodal  tensors to something like multimodal train inputs and still calling the orginal field as multimodal inputs.

zhuzilin (approved):

PR #1283, opened 2025-12-28
fix: remove invalid None default and fix misleading underscore variable naming

lancerts:
## Summary

Fix two minor code issues:

1. **`metric_utils.py`**: Remove invalid `None` default from `has_repetition(text: str = None)` which would cause `TypeError` when called without arguments
2. **`seqlen_balancing.py`**: Rename `_i` to `i` since the variable is used in the error message (underscore prefix convention implies unused)


lancerts:
@yitianlian ty

PR #1285, opened 2025-12-29
add bshd support

yueming-yuan:
https://github.com/radixark/miles/pull/341

### DIFF
--- a/slime/backends/megatron_utils/data.py
+++ b/slime/backends/megatron_utils/data.py
@@ -111,7 +111,7 @@ def get_batch(
     ):
         prompt_length = total_length - response_length
         loss_mask = F.pad(loss_mask, (prompt_length - 1, 1), value=0)
-        loss_mask = slice_with_cp(loss_mask, 0, qkv_format, max_seq_len=max_seqlen)
+        loss_mask = slice_with_cp(loss_mask, 0, qkv_format, max_seqlen)
         loss_masks.append(loss_mask)
 
     if qkv_format == "bshd":


slime/utils/arguments.py
```diff
@@ -1639,6 +1646,16 @@ def slime_validate_args(args):
         args.prefill_num_servers is not None and args.rollout_external
     ), "prefill_num_servers cannot be set when rollout_external is set."
 
+    assert args.qkv_format in [
+        "thd",
+        "bshd",
+    ], f"qkv_format {args.qkv_format} is not supported. (only 'thd' and 'bshd' are supported)"
```
zhuzilin: this check doesn't seem necessary as the argparse choices should already do the check

slime/backends/megatron_utils/loss.py
```diff
@@ -543,7 +570,12 @@ def policy_loss_function(
         # [decouple IS and rejection] Rebuild sum_of_sample_mean with modified_response_masks for denominator correction
         # modified_response_masks will be sliced with cp in get_sum_of_sample_mean
         sum_of_sample_mean = get_sum_of_sample_mean(
-            total_lengths, response_lengths, modified_response_masks, args.calculate_per_token_loss
+            total_lengths,
+            response_lengths,
+            modified_response_masks,
+            args.calculate_per_token_loss,
+            args.qkv_format,
+            batch.get("max_seq_lens", None),
```
zhuzilin:
```suggestion
            batch["max_seq_lens"],
```
the batch will always have "max_seq_lens" if it is in the `keys` passed to `get_batch`. I recommend using `batch["max_seq_lens"]` to prevent key name mismatch.

slime/utils/arguments.py
yueming-yuan: good catch! deleted

slime/backends/megatron_utils/loss.py
yueming-yuan: modified. thanks~

PR #1298, opened 2025-12-31
[data][feat] add large dataset support

SwordFaith:
Thank you for the efforts on https://github.com/radixark/miles/pull/246 and the hard work contributed by @Ratish1!


ChangyiYang:
@codex review

ChangyiYang:
@codex review

slime/backends/fsdp_utils/actor.py
```diff
@@ -712,7 +712,22 @@ def _has_rollout_log_probs(batch) -> bool:
...
 
-        loss = pg_loss - self.args.entropy_coef * entropy_loss
+        # Compute loss based on loss_type
+        if self.args.loss_type == "sft_loss":
+            # SFT loss: simple negative log-likelihood
+            sft_loss = -log_probs
+            if self.args.calculate_per_token_loss:
+                sft_loss = sum_of_token(sft_loss, response_lengths, loss_masks)
+            else:
+                sft_loss = sum_of_sample_mean(sft_loss, response_lengths, loss_masks)
```
PopSoda2002: I think we just need to `calculate_per_sample_loss` in SFT?

PopSoda2002 (commented):
Thanks for your great work! I am not sure if we need to seperate `datasets` and `SFT` into different PRs, currently this PR seems too large for reviewing

slime/rollout/data_source.py
```diff
@@ -52,52 +52,157 @@ def __init__(self, args):
...
+                max_length=self.args.rollout_max_prompt_len,
+                prompt_key=self.args.input_key,
+                label_key=self.args.label_key,
+                tool_key=self.args.tool_key,
+                metadata_key=self.args.metadata_key,
+                multimodal_keys=self.args.multimodal_keys,
+                seed=self.args.rollout_seed,
+                apply_chat_template=self.args.apply_chat_template,
+                apply_chat_template_kwargs=self.args.apply_chat_template_kwargs,
+                dp_size=self._dp_size or 1,
```
zhuzilin: We should not pass dp size to the data source because the data source is used in rollout manager, which does not have dp ranks.

SwordFaith:
> Thanks for your great work! I am not sure if we need to seperate `datasets` and `SFT` into different PRs, currently this PR seems too large for reviewing

I’ll work on some simplifications with @ChangyiYang today. Afterward, could you review it again and discuss if we should split this PR into two?

slime/rollout/data_source.py
SwordFaith: > We should not pass dp size to the data source because the data source is used in rollout manager, which does not have dp ranks.

Apologies for the mistake, I’ll address it in the upcoming commits.

PopSoda2002:
> > Thanks for your great work! I am not sure if we need to seperate `datasets` and `SFT` into different PRs, currently this PR seems too large for reviewing
> 
> I’ll work on some simplifications with @ChangyiYang today. Afterward, could you review it again and discuss if we should split this PR into two?

Yeah sure, definitely, always willing for help

slime/backends/fsdp_utils/actor.py
ChangyiYang: After further discussion, we realized that SFT should only use per-token loss.
We’ll simplify this logic by keeping token loss only here.
For users who still try to use sequence / per-sample loss in SFT, we’ll explicitly raise an error to avoid silent misconfiguration.

PR #1306, opened 2026-01-02
Set base_gpu_id for sglang from placement groups

vpj:
Current the SGLang base_gpu_id is recalculated without using the actual placement group gpu id. This makes slime not work with any other jobs on the same ray cluster if they are sharing the same node. This code keeps track of the gpu ids of the placement groups and pass it for sglang creation.


zhuzilin (approved):

zhuzilin:
This is really nice, thank you so much!

PR #1313, opened 2026-01-03
[docker] Fix sglang ima on mtp + pd disaggregation

zhuzilin:


yudian0504:
Hi @zhuzilin zhuzilin  , we’re also suffering from the same IMA issue. The difference is that in our case it’s a non‑PD scenario. Similarly, I’ve observed abnormal values in topk_index, but I haven’t yet figured out which ones were incorrectly written there. I noticed that you added a clamp operation here, so I’d like to ask whether you know the root cause of these abnormal values.

zhuzilin:
@yudian0504 I haven't found the root cause either, and this is a temperary work around.

PR #1333, opened 2026-01-05
[Feature] Option not to save optimizer states to save disk space

yzlnew:
Add an option to disable optimizer states in ckpt to save disk space / reduce OOM issue during saving.


zhuzilin:
We can use `--no-save-optim` in megatron to achieve this.

yzlnew:
> We can use `--no-save-optim` in megatron to achieve this.

I see, renaming the option name to `--no-save-optim`.

PR #1344, opened 2026-01-06
[Fix] Update deprecated sglang ep args in docs and scripts

coding-famer:
update `enable-ep-moe`, `enable-deepep-moe` to `ep-size`, `moe-a2a-backend deepep` in both docs and scripts


coding-famer:
and blog remains unchanged

PR #1346, opened 2026-01-06
[Bug] Log metadata after loading from state_dict

ChangyiYang:
Previous log print old metadata, which seems wrong.
Move log after the metadata update.


ChangyiYang:
data source's meta data is actually deprecated. will raise a pr to remove metadata entirely

PR #1355, opened 2026-01-07
Feat: multi-threads data fetching for sft data

UbeCc:


yitianlian:
You can use train_async.py for sft training

PR #1360, opened 2026-01-08
load dummy sglang

lilei199908:


lilei199908:
we can add sglang-load-format dummy to do it.

PR #1364, opened 2026-01-08
Add OSWorld VLM training cookbook and integration

jbarnes850:
This PR adds an [OSWorld ](https://github.com/xlang-ai/OSWorld)VLM training cookbook and reference integration for multi‑turn desktop automation in Slime. It is written for a reviewer who has not seen OSWorld before and explains how Slime connects to a VM‑backed GUI environment, why an HTTP bridge is required, and how training is run end‑to‑end without touching Slime core code.

architecture

OSWorld depends on desktop‑env (torch 2.5.1), while Slime’s training stack uses torch 2.9+, so the environment runs on the host and training runs in the container. 

The cookbook makes that split explicit, proxies environment calls through a stateless HTTP server, and drives a multi‑turn VLM rollout loop that produces Qwen3‑VL multimodal tokens and aligned loss masks. Experience replay is used to avoid advantage collapse in sparse reward tasks, and reward shaping provides partial credit for valid actions, execution, and UI changes.

Reproducible artifacts are referenced to validate the pipeline without hunting for external dependencies: the task registry and replay buffer are in `Jarrodbarnes/osworld-union-v1`, the SFT warmup checkpoint is `Jarrodbarnes/osworld-vlm-sft-step25`, and the GSPO checkpoint is `Jarrodbarnes/osworld-vlm-gspo-step25`. Metrics for the training runs are tracked in W&B at `jbarnes850-near-protocol/osworld-grpo`.

The README documents the OSWorld subset used for training (Ubuntu only), how to launch the host VM, how the HTTP bridge is wired, and how to run training. Verification was done by running pre‑commit and compiling the example module.

This PR keeps all code confined to `examples/osworld/` and does not modify Slime core paths.

### DIFF
--- a/examples/osworld/train_sft.sh
+++ b/examples/osworld/train_sft.sh
@@ -71,7 +71,6 @@ SFT_ARGS=(
     --rollout-function-path slime.rollout.sft_rollout.generate_rollout
     --prompt-data "${DATASET_LOCAL}/osworld_reasoning_sft_v1.jsonl"
     --input-key messages
-    --apply-chat-template
     --rollout-shuffle
 
     # Training schedule - conservative for small dataset


GuanxingLu:
This looks awesome. Any chance you could share some training logs (e.g., reward curve) or a quick demo?

zhuzilin:
I'm afraid I have to close this PR because this is not suitable as an example of slime, which should be simple and show some specific features of slime.
I would recommend to turn this PR into an independent repo and we'd love to add link to the repo in the readme.

PR #1365, opened 2026-01-08
Add return_exceptions=True to asyncio.gather for partial success

JackXu0:
- Add return_exceptions=True to asyncio.gather in generate_and_rm_group
- Handle exceptions by marking failed samples as ABORTED
- Allows partial success when some samples in a group fail


ZhuochengXu:
Seems it is by design to fail fast on permanent errors.

PR #1367, opened 2026-01-09
[FSDP] Add argument validation for FA3 with cp

Beichen-Ma:
Ring attention implementation does not support Flash Attention 3 yet. Using --attn-implementation flash_attention_3 with --context-parallel-size > 1 would silently cause NaN loss during training. This change adds an early validation check
that raises a clear error with actionable guidance.


zhuzilin:
The cp for fsdp will be redesigned, close for now.

PR #1372, opened 2026-01-09
Fix: Apply loss mask to KL in REINFORCE++ returns calculation

### LINKED ISSUE #1370: 【BUG】no masking for tool_response in token_level_rewards when applied to agentic rl
In slime/utils/ppo_utils.py def get_reinforce_plus_plus_returns(...)

when **calculating 'returns_for_seq' with 'token_level_rewards', no mask is applied to the kl w.r.t. tool_response**, this may be problematic when applied to agentic rl, because tool_response is returned by the environment, and should not engage in the forward computation. Otherwise, all kl of tool_response may accumulate when calculating the return.

Do I understand it correctly?

kaysonyu:
## Summary

  Fixes #1370 - Apply loss mask to KL divergence before computing returns in get_reinforce_plus_plus_returns.

  ## Problem

  In REINFORCE++ advantage estimation, the KL divergence values were not masked before computing returns. This caused environment-generated tokens (e.g., tool_response in agentic RL scenarios) to incorrectly contribute to the KL penalty that propagates through the cumulative return calculation.
  
   ## Solution

  Apply loss_mask to zero out KL values at masked positions before computing returns:

```python
  # Before:
  token_level_rewards = -kl_coef * full_kl_response

  # After:
  masked_kl = full_kl_response * full_mask
  token_level_rewards = -kl_coef * masked_kl
  ```


zhuzilin:
close as 1370 doesn't need to be fixed.

zhuzilin:
sorry I was wrong about the issue. thank you for the fix!

PR #1377, opened 2026-01-10
Fix: fix double-trim bug in entropy computation for last CP rank

Beichen-Ma:
In `get_logprob_and_entropy_with_cp`, the logits is trimed twice for last rank, causing missing entropy for one token position and incorrect entropy values when using CP with fsdp backend.

Consolidate into single shifted_logits variable used for both log_probs and entropy.


yueming-yuan:
(this might not be compatible with new FSDP impl...)

PopSoda2002:
I believe it's not a bug, we can sync for this although the implementation is a little bit hardcode

Issue #1062, opened 2025-12-08
[FSDP] Try to optimize context parallelism


PopSoda2002:
Currently the context parallelism's training time is like 2 times when we use `cp=2` than `cp=1`, we should try to optimize this for long sequence, maybe reduce the `all gather entropy and logp` and calculate the local loss to reduce the numbers of communication

cklxx:
I’m not sure I can fully complete this, but I’d like to start with a minimal experiment to validate the bottleneck and share the findings.

PopSoda2002:
yeah thanks! Let's go!

embersax:
Hi all, how is the progress on this issue? I'm new to Slime and also trying to contribute. Can I also be assigned to this?

HaZ-K:
Hi all, how is the progress on this issue? I'm new to Slime and also trying to contribute. Can I also be assigned to this?

Riyasushin:
Hi all,  I'm new to Slime and also trying to contribute. How should I start helping

fmh66:
Hi everyone, I'm new to Slime and I'd like to contribute as well. How should I get started helping out?

PR #1383, opened 2026-01-12
[FSDP] Optimize context parallel

Beichen-Ma:
# Motivation
Try to optimize context parallel for FSDP backend #1062 by eliminating the `all_gather` for entropy and log_probs. Instead of gathering log_probs and entropy across all CP ranks and then distributing them, each rank now computes and retains only its local portian, with handling during unpacking and loss computation.

# Changes
- `get_logprob_and_entropy_with_cp` change from all_gather full log_probs/entropy to return local shards only
- `unpack_sequences_with_cp` handle tensor slicing, extracting only the portion of each sequence's response that overlaps with the current rank
- `all_reduce` the final loss across cp ranks

# Impact
During training, `get_logprob_and_entropy_with_cp` is called 3 times to obtain ref logprobs, actor logprobs and current logprobs. Three `all_gather` operations on full-lengths response log_probs+entropy replaced by 1 scalar `all_reduce`.

# Concerns
- The `sum_of_sample_mean` is now computed as a sum of local contributions.

### DIFF
--- a/docs/en/get_started/customization.md
+++ b/docs/en/get_started/customization.md
@@ -406,4 +406,14 @@ def custom_hook(args, rollout_id, step_id, model, optimizer, opt_param_scheduler
 - Custom routing logic
 - Caching and optimization
 
+---
+
+### 19. MoE Routing Replay
+
+Stabilize MoE RL training by recording and replaying expert routing decisions to ensure consistency.
+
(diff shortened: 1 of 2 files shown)


yueming-yuan:
(this might not be compatible with new FSDP impl...)

PR #1389, opened 2026-01-12
Add support for per token reward/advantages with a custom_reward_post_process_path

vpj:
This pull request lets you set per token rewards with a custom_reward_post_process_path


vpj:
@zhuzilin

PR #1438, opened 2026-01-17
[Fix] Return origin_samples instead of False in filter_long_prompt

kaysonyu:
## Summary

  Fix filter_long_prompt to return origin_samples instead of False when skipping length filtering. The function's return type is list[Sample], but it was returning a boolean in certain code paths.

  ## Problem

  The filter_long_prompt function had two code paths that returned False instead of the expected list[Sample]:

  1. When max_length is None: Returned False (though this branch is currently unreachable due to caller-side check)
  2. When prompt is not a string (e.g., pre-formatted chat template list): Returned False


zhuzilin:
nice catch! thx!

PR #1444, opened 2026-01-17
[fsdp] remove tis

zhuzilin:


sordonia:
hi ! may I ask why removing tis from fsdp? Thank you!!!

PR #1461, opened 2026-01-20
fix: correct inverted condition for NaN check in loss and grad

JackXu0:
The condition `check_for_nan_in_loss_and_grad` was incorrectly negated, causing the NaN check to be skipped when it should run and vice versa.


zhuzilin:
The reason for setting `check_for_nan_in_loss_and_grad` to False is that we can skip the nan and inf check in megatron, so the origin impl is intended.

ZhuochengXu:
I see. Closing. Thanks for the context.

PR #1470, opened 2026-01-20
fix: validate reward_key when RM returns dict

zpqiu:
When RM (e.g., dapo) returns a dict instead of a scalar value, raise a clear error if neither reward_key nor eval_reward_key is set.

This prevents confusing errors like:
```
  TypeError: unsupported operand type(s) for +: 'int' and 'dict'
```

The error message now includes available keys and suggests the fix:
```
  Please specify --reward_key or --eval_reward_key (e.g., --reward_key score)
```


zpqiu:
@zhuzilin  could you take a look, thank you

PR #1491, opened 2026-01-25
[Fix] support converting torch_dist to hf for qwen3vl dense model

p1k0pan:
After training Qwen3-VL-8B with Megatron, it was unable to convert torch_dist to hf. Adding convert code. 
Tested on Qwen3-VL-8B, not sure whether suitable for Qwen3-VL moe model.


weixiao-zhan:
Hi, the file name might need to be `qwen3_vl.py`. 

```
(head, rank=0, pid=40184)     from .qwen3_vl import convert_qwen3vl_to_hf
(head, rank=0, pid=40184) ModuleNotFoundError: No module named 'slime.backends.megatron_utils.megatron_to_hf.qwen3_vl'
```

PR #1495, opened 2026-01-26
fix: fp16 config and grad scaling

yxyOo:
## Summary
Ensure FP16 training is consistent and safe by setting required optimizer params, aligning fp16/bf16 flags, and avoiding duplicate grad scaling.

## Changes
- Set required FP16 optimizer parameters in code to avoid missing user‑configurable CLI options.
- Set `store_param_remainders=false` to reduce memory usage, since this option isn’t exposed by Megatron’s public config.
- Skip duplicate grad scaling checks during FP16 training.
-  Refs: https://github.com/volcengine/verl/pull/4086

## Results
fp16 shows faster convergence speed and lower mismatch between training and rollout.
all

### DIFF
--- a/slime/backends/megatron_utils/model.py
+++ b/slime/backends/megatron_utils/model.py
@@ -435,7 +435,7 @@ def train_one_step(
     )
 
     valid_step = True
-    # FP16 mode does not perform NaN validation on loss and gradients, as this validation process would result in duplicate gradient scaling (grad scale).
+    # FP16 mode does not perform NaN validation on loss and gradients, as this validation process would result in duplicate gradient scaling.
     if not getattr(args, "check_for_nan_in_loss_and_grad", True) and not args.fp16:
         found_inf_flag = optimizer.prepare_grads()
         if found_inf_flag:


slime/backends/megatron_utils/model.py
```diff
@@ -112,6 +112,15 @@ def setup_model_and_optimizer(
...
             kwargs[f.name] = getattr(args, f.name)
+    if args.fp16:
+        kwargs["bf16"] = False
+        kwargs["fp16"] = True
+        kwargs["params_dtype"] = torch.float16
+        kwargs["initial_loss_scale"] = 32768
+        kwargs["min_loss_scale"] = 1
+        kwargs["use_precision_aware_optimizer"] = True
+        kwargs["store_param_remainders"] = False
+        logger.info(f"FP16 mode enabled. Optimizer config: {kwargs}")
```
zhuzilin, line 123: I think that these params can be set with the
```bash
    for f in dataclasses.fields(OptimizerConfig):
        if hasattr(args, f.name):
            kwargs[f.name] = getattr(args, f.name)
```
For example, you can set the `initial_loss_scale` with `--initial-loss-scale 32768`.

yxyOo: Yes, configurable, but these params must be enabled with FP16—otherwise performance degrades and users may miss them.

yxyOo: And store_param_remainders cannot be set this way (not exposed in Megatron’s public config), while other params are permissible.
image

PR #1499, opened 2026-01-27
[Doc] Add doc for slime router

Hecate0821:
Add docs for slime router, including radix tree and R3 use case.


Hecate0821:
The pre commit error seems not caused by this doc change

PR #1516, opened 2026-01-29
[Multimodal] make multimodal processing robust

coding-famer:
**Modifications:**
1. Use explicit base64 encoding.
2. Force `return_tensors` to None and set `return_tensors='pt'` for multimodal inputs.
3. Lazy import `qwen_vl_utils` when loading processor.

### DIFF
--- a/slime/utils/processing_utils.py
+++ b/slime/utils/processing_utils.py
@@ -11,8 +11,6 @@ logger = logging.getLogger(__name__)
 # Reference: https://github.com/QwenLM/Qwen3-VL/blob/main/qwen-vl-utils/README.md
 DEFAULT_PATCH_SIZE = 14
 
-_qwen_process_vision_info = None
-
 
 def load_tokenizer(name_or_path: str, **kwargs):
     return AutoTokenizer.from_pretrained(name_or_path, **kwargs)
@@ -23,8 +21,6 @@ def build_processor_kwargs(multimodal_inputs: dict | None = None) -> dict:
     forced = {
         # force return_tensors to None for input_ids
         "return_tensors": None,
-        # have been resized by qwen_vl_utils, update this when supporting other models
-        "do_resize": False,
     }
     modality_forced = {"return_tensors": "pt"}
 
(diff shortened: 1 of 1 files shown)


slime/utils/processing_utils.py
```diff
@@ -27,19 +53,30 @@ def load_processor(name_or_path: str, **kwargs):
...
+
     return proc
 
 
 def process_vision_info(prompt, processor):
-    # temporary solution, will write image utils for slime later
-    from qwen_vl_utils import process_vision_info
+    # TODO: temporary solution, will write image utils for slime later
+    if _qwen_process_vision_info is None:
+        raise ImportError("qwen_vl_utils is not installed. Install it with: pip install qwen-vl-utils")
```
zhuzilin: hmm... I don't get why we need to move the import to the function above...

slime/utils/processing_utils.py
```diff
@@ -50,4 +87,5 @@ def encode_image_for_rollout_engine(image) -> str:
     if image.mode != "RGB":
         image = image.convert("RGB")
     image.save(buffer, format="PNG")
-    return base64.b64encode(buffer.getvalue()).decode("utf-8")
+    image_base64 = base64.b64encode(buffer.getvalue()).decode("utf-8")
+    return f"data:image/png;base64,{image_base64}"
```
zhuzilin, line 76: shall we move the `f"data:image/png;base64,{image_base64}"` template into `sglang_rollout.py`? It seems like a template that is tightly connect to http payload.

coding-famer: I was thinking about potential future modalities (audio, video, etc.) that may have different MIME types. Keeping the data formatting in each encode functions make `sglang_rollout.py` doesn't need to handle different MIME types for each modality.
(Although SGLang actually just matches `data:` and `,` without parsing the MIME type, but including it makes the format less confusing.）

slime/utils/processing_utils.py
```diff
@@ -11,11 +11,37 @@
...
     return AutoTokenizer.from_pretrained(name_or_path, **kwargs)
 
 
+def build_processor_kwargs(multimodal_inputs: dict | None = None) -> dict:
+
+    forced = {
+        # force return_tensors to None for input_ids
+        "return_tensors": None,
+        # have been resized by qwen_vl_utils, update this when supporting other models
+        "do_resize": False,
```
coding-famer: Removing this for now. Since SGLang re-processes images internally and doesn't expose a `do_resize` option.

slime/utils/processing_utils.py
coding-famer: updated

Issue #1498, opened 2026-01-27
Creating a new Dataset source throws in `rollout.py` because `len(data_source.dataset)` is used.


TSunny007:
I'm trying to create an abstract dataset 
```

import json
import logging
from slime.rollout.data_source import DataSource
from slime.utils.processing_utils import load_processor, load_tokenizer
from slime.utils.types import Sample

logger = logging.getLogger(__name__)


class SFTDataSource(DataSource):
    def __init__(self, args):
        self.args = args
        self.tokenizer = load_tokenizer(args.hf_checkpoint, trust_remote_code=True)
        self.processor = load_processor(args.hf_checkpoint, trust_remote_code=True)
        self.file = args.prompt_data
        if not self.file:
            raise ValueError("prompt_data argument is required for SFTDataSource")
        self.stream = open(self.args.prompt_data)


    
    def get_samples(self, num_samples: int) -> list[list[Sample]]:
        """Return num_samples samples"""
        samples = []
        while len(samples) < num_samples:
            line = self.stream.readline()
            
            # EOF - reset to beginning
            if not line:
                self.stream.seek(0)
            
            data = json.loads(line)
            if self.processor:
                from slime.utils.processing_utils import process_vision_info

                assert isinstance(
                    data, list
                ), f"prompt must be a list when processor is not None, got {type(data)} instead"
                multimodal_inputs = process_vision_info(data, self.processor)

            metadata = data.get("metadata", {})
            if "id" in data:
                metadata["id"] = data["id"]
            
            samples.append([Sample(
                prompt=data,
                metadata=metadata,
                multimodal_inputs=multimodal_inputs,
            )])
        
        return samples

        
    def add_samples(self, samples: list[list[Sample]]):
        """Add samples back to the data source"""
        pass
        
    def save(self, rollout_id):
        """Save state for checkpointing"""
        pass
        
    def load(self, rollout_id=None):
        """Load state from checkpoint"""
        pass
```

The content of this loader isn't what's important - it'll be expanded to support my use cases very soon.

What is throwing is the following:
```
---------------------------------------
Job 'raysubmit_bnqeJDFmQkNanYpK' failed
---------------------------------------

Status message: Job entrypoint command failed with exit code 1, last available logs (truncated to 20,000 chars):
    raise value.as_instanceof_cause()
[...]

PR #1518, opened 2026-01-29
Make `DataSource` implement `__len__` to standardize the data source contract

TSunny007:
Fixes issue [#1498](https://github.com/THUDM/slime/issues/1498)

### DIFF
--- a/slime/rollout/data_source.py
+++ b/slime/rollout/data_source.py
@@ -38,7 +38,7 @@ class DataSource(abc.ABC):
         """
         Load the state of the data source
         """
-    
+
     @abc.abstractmethod
     def __len__(self) -> int:
         """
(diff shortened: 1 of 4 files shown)


slime/rollout/data_source.py
```diff
@@ -207,7 +216,7 @@ def update_metadata(self, metadata: dict):
     def get_metadata(self):
         return self.metadata
 
-    def get_buffer_length(self):
+    def __len__(self) -> int:
         return len(self.buffer)
```
zhuzilin, line 220: The `get_buffer_length` is for the size of the buffer instead of the whole dataset, we can still inherit the `__len__` from `RolloutDataSource`

TSunny007: Gotcha! Maybe its good to have a mixin down the line for buffered datasets but that can be added later

TSunny007:
@zhuzilin could you take another look at this when you have the time please? we are working on supporting streaming sft

Issue #1523, opened 2026-01-30
OOM during long-sequence RL training caused by entropy backward tensors


huanglei-lpd:
# Issue: OOM during long-sequence RL training caused by entropy backward tensors

## Problem Description

When training with **long sequences** (total sample length >30k tokens), CUDA OOM occurs during the training loop. The OOM happens on the last pipeline parallel stage, for **calculating entrophy loss**.

Even with CP & log_probs_chunk, and reducing the seq-len,  Megatron should got plenty free mem, OOM still occurs at final pipeline stage.

## Environment

- Hardware: Multi-node setup with 8 GPUs per node (H200 139GB)
- Parallelism: TP=4, PP=4, CP=2
- Key parameters:
   - `--max-tokens-per-gpu 32768`
   - `--log-probs-chunk-size 8192`
   - `--entropy-coef 0.00`

## Error Log

```text
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.14 GiB. 
GPU 3 has a total capacity of 139.81 GiB of which 1.50 GiB is free.

File "slime/backends/megatron_utils/loss.py", line 156, in get_log_probs_and_entropy
    log_prob, entropy = calculate_log_probs_and_entropy(...)
File "slime/utils/ppo_utils.py", line 668, in calculate_log_probs_and_entropy
    entropy = compute_entropy_from_logits(logits_chunk.clone(), tp_group)
File "slime/utils/ppo_utils.py", line 173, in forward
    normalized_vocab_parallel_logits = vocab_parallel_logits - logits_max
```

### Entropy backward tensors consume significant memory

The [`_VocabParallelEntropy.forward` method](https://github.com/THUDM/slime/blob/v0.2.2/slime/utils/ppo_utils.py#L181) saves three tensors via ctx.save_for_backward:
- `vocab_parallel_logits`
- `softmax_logits`
- `sum_softmax_times_logits`

For long sequences with large vocabulary, these tensors are substantial:
- Shape: (seq_len, vocab_size/TP), here chunk has no contribution for memory reduce.
- Memory per sample: ~5-6 GB just for backward intermediate tensors, even more huge when expand to 128k.

### Unnecessary computation when entropy_coef=0

The current implementation always computes entropy with gradient tracking (with_entropy=True is hardcoded), even when entropy_coef=0.
[...]

PR #1524, opened 2026-01-30
fix: resolve OOM in long-sequence training via conditional entropy gradient tracking

ppraneth:
This PR addresses the CUDA Out of Memory (OOM) issues #1523  encountered during training with long sequences (e.g., >30k tokens) by implementing **conditional gradient tracking** for the entropy term.

#### **The Problem**

In the previous implementation, the entropy calculation was always differentiable, regardless of the `entropy_coef` value. To support the backward pass, the system was forced to store massive intermediate activation tensors (logits and softmax outputs) with shapes of `(seq_len, vocab_size / TP)`. For long sequences and large vocabularies, these tensors consumed **5–6 GB of VRAM per sample** on the last pipeline stage, leading to OOM even when the entropy contribution to the loss was zero.

#### **The Solution**

I have decoupled entropy for monitoring (logging) from entropy for training (loss). 

* **Conditional Logic**: Gradients for the entropy term are now only tracked if `args.entropy_coef > 0`.
* **Memory Recovery**: When `entropy_coef` is 0.0, entropy for logging is computed within a `torch.no_grad()` context. This prevents PyTorch from allocating memory for backward tensors, effectively reclaiming several gigabytes of VRAM per sequence.
* **Backend Support**: This fix has been applied to both the **Megatron** and **FSDP** backends to ensure consistency across configurations.

#### **Files Changed**

* `slime/utils/ppo_utils.py`: Added `requires_entropy_grad` flag and `no_grad` context to `calculate_log_probs_and_entropy`.
* `slime/backends/megatron_utils/loss.py`: Passed gradient requirement flag based on `entropy_coef`.
* `slime/backends/fsdp_utils/actor.py`: Implemented conditional tracking and added `contextlib` import.

### DIFF
--- a/slime/backends/fsdp_utils/actor.py
+++ b/slime/backends/fsdp_utils/actor.py
@@ -1,3 +1,4 @@
+import contextlib
 import logging
 import os
 import random
@@ -342,6 +343,7 @@ class FSDPTrainRayActor(TrainRayActor):
                         target_tokens=batch["tokens"],
                         allow_compile=not self.args.true_on_policy_mode,
                         temperature=self.args.rollout_temperature,
+                        requires_entropy_grad=False,
                     )
                     batch[f"{store_prefix}log_probs"] = log_probs_result
                     if store_prefix == "":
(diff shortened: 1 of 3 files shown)


lilei199908:
thanks, we are refactoring this part for better menmory use.

ppraneth:
@lilei199908 could you review this PR and suggest any edits? I believe this should reduce memory usage.

lilei199908:
> @lilei199908 could you review this PR and suggest any edits? I believe this should reduce memory usage.

LGTM, we will merged it soon! thanks

ppraneth:
@zhuzilin Can you check and merge this pr?

ppraneth:
@yitianlian Can you check and merge this pr?

PR #1535, opened 2026-02-02
[bug] fix: gracefully handle datapoints with no multimodal input in a multimodal dataset

hleehlee-amazon:
Previously we assume each entry in image dataset contains image data, it is not the case for mixed dataset. This PR allows the data loader to handle mixed datasets where some datapoints have multimodal inputs and some don't.

Successfully ran an SFT job in a local environment with the change.

### DIFF
--- a/slime/utils/data.py
+++ b/slime/utils/data.py
@@ -128,7 +128,9 @@ def _build_messages(data: dict, prompt_key: str, as_conversation: bool, multimod
         for type_name, data_key in multimodal_keys.items():
             mt = MultimodalTypes.get(type_name)
             if mt:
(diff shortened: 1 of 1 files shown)


hleehlee-amazon:
@zhuzilin Could you please review this?

coding-famer:
Does this part also need to be updated?
https://github.com/THUDM/slime/blob/58ee74898f72aacd2413ffdf7685dc0eee72d8d8/slime/ray/rollout.py#L388-L389

hleehlee-amazon:
@coding-famer Looks like the `None`s in rollout.py is handled already:

https://github.com/THUDM/slime/blob/main/slime/backends/megatron_utils/actor.py#L200-L201

PR #1538, opened 2026-02-03
[Feature] Add megatron version for on policy distillation

yitianlian:


slime/backends/megatron_utils/loss.py
```diff
@@ -310,25 +350,17 @@ def compute_advantages_and_returns(args: Namespace, rollout_data: RolloutBatch)
...
-            teacher_log_prob - student_log_prob
-            for teacher_log_prob, student_log_prob in zip(teacher_log_probs, student_log_probs, strict=False)
-        ]
-        returns = advantages
-
     else:
         raise NotImplementedError(f"advantage_estimator {args.advantage_estimator} is not supported. ")
 
+    # Apply on-policy distillation KL penalty to advantages (orthogonal to advantage estimator)
+    apply_opd_kl_to_advantages(
```
zhuzilin: It will be better if we move the early return in this function to outside, similar to:
```python
if not args.use_opd:
    apply_opd_kl_to_advantages(...)
```

slime/utils/arguments.py
```diff
@@ -730,6 +730,42 @@ def add_algo_arguments(parser):
...
+                type=str,
+                default=None,
+                help=(
+                    "The checkpoint for OPD teacher model. Required when --opd-type=megatron. "
+                    "The teacher model should have the same architecture as policy/ref model."
+                ),
+            )
+            parser.add_argument(
+                "--opd-teacher-ckpt-step", type=int, default=None, help="The checkpoint step for OPD teacher model."
+            )
```
zhuzilin: we can have a new `add_on_policy_distillation_arguments` function to manage the arguments for opd.

PR #1550, opened 2026-02-05
[bug] fix: fix type hint

hleehlee-amazon:
Fix type hint in Sample class


hleehlee-amazon:
@zhuzilin Could you please review this?

PR #1551, opened 2026-02-05
[bug] fix: fix type hint in `Sample` class

hleehlee-amazon:
Fixed type hint in `Sample` class attributes `multimodal_inputs` and `multimodal_train_inputs` to allow for `None`.


hleehlee-amazon:
@zhuzilin Could you please take a look? cc @TSunny007 @nanjiangwill

PR #1592, opened 2026-02-17
[Fix] Minor fix for properly finishing / flushing wandb logging metrics at exit

silunw:
## Summary
This fix makes W&B logging shutdown deterministic so wandb metrics are not lost at job end, especially in Ray-based execution.
- Add `finish_tracking(args)` in `slime/slime/utils/logging_utils.py` to explicitly call `wandb.finish()` when W&B is enabled.
- Call `finish_tracking(self.args)` in `RolloutManager.dispose()` (`slime/slime/ray/rollout.py`) to flush metrics from the rollout actor process (**secondary** W&B run).
- Call `finish_tracking(args)` at the end of `train()` (`slime/train.py`) to flush the driver process metrics (**primary** W&B run).
- Keep behavior unchanged during training; only shutdown/flush behavior is updated.

## Why this change
SLIME initializes W&B in both the driver and the rollout Ray actor. Relying on process teardown can drop buffered metrics, which sometimes causes missing eval metrics on dashboards after training ends.

## Expected impact
Eval metrics reliably appear in W&B after job completion.
No change to metric computation logic, rollout flow, or training semantics.


silunw:
Hi @zhuzilin , I'm unable to assign reviewers myself. Would you be able to take a look at this minor fix? Thanks!

PR #1596, opened 2026-02-18
[Feature] Add configurable arguments for rollout manager actor

TSunny007:
Add ability to configure the resources for RolloutManager if necessary. 

The default `num_cpus=1` is too restrictive in some use cases and it'd be beneficial to be able to override these parameters if necessary. According to ray [docs](https://docs.ray.io/en/latest/ray-core/api/doc/ray.actor.ActorClass.options.html#ray.actor.ActorClass.options), sending `None` values to the other parameters will be equivalent to the defaults ray chooses.


TSunny007:
@zhuzilin what do you think about this refactor? I'm observing that SFT can be bottlenecked because of the default resources allocated for the rollout manager

PR #1623, opened 2026-02-25
feat: add --lazy-multimodal-load to defer image process to rollout time

yzlnew:
Avoid OOM during Dataset init for large VLM datasets by deferring process_vision_info calls to rollout time. Controlled by the --lazy-multimodal-load flag (default off, preserving existing behavior).

### DIFF
--- a/slime/utils/types.py
+++ b/slime/utils/types.py
@@ -13,6 +13,7 @@ class Sample:
     index: int | None = None
     # prompt
     prompt: str | list[dict[str, str]] = ""
+    raw_prompt: list[dict[str, str]] | None = None  # original conversation-format prompt for lazy multimodal loading
     tokens: list[int] = field(default_factory=list)
     multimodal_inputs: dict[str, Any] | None = None  # raw multimodal data, e.g. images, videos, etc.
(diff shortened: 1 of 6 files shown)


slime/utils/types.py
```diff
@@ -13,6 +13,7 @@ class Sample:
     index: int | None = None
     # prompt
     prompt: str | list[dict[str, str]] = ""
+    raw_prompt: list[dict[str, str]] | None = None  # original conversation-format prompt for lazy multimodal loading
```
zhuzilin, line 16: can we reuse the prompt field to store the origin conversation-format prompt?

yzlnew: when passing `--apply-chat-template`, the original format is converted to string?

coding-famer:
Can I ask about the progress of this PR? In fact, we don't need conversation-format prompts in rollout side. @yzlnew Would you mind if I continue from this PR and open a new one?

yzlnew:
> Can I ask about the progress of this PR? In fact, we don't need conversation-format prompts in rollout side. @yzlnew Would you mind if I continue from this PR and open a new one?

Feel free to continue from this PR and open a new one. I'm closing this pr btw.

PR #1662, opened 2026-03-03
[fix]: Qwen3.5-35B-A3B 8-GPU: set TP size to 2 for num_query_groups=2

none0663:
- TP: Megatron needs num_query_groups % tensor_model_parallel_size == 0. Model has 2 query groups, so set --tensor-model-parallel-size 2 (with --expert-model-parallel-size 8) for single-node 8-GPU.
Clipboard_Screenshot_1772591767

https://github.com/THUDM/slime/blob/d78bb43eed5d714d681271fcf8c04f0f28f6b67c/scripts/models/qwen3.5-35B-A3B.sh#L23

- data.py / rollout: For multimodal models, use the processor only when a sample has real multimodal content (multimodal_inputs with non-None values). For text-only samples, use batched tokenizer(prompts, add_special_tokens=False) in filter_long_prompt and tokenizer in rollout so pure-text data is not run through the processor.

Just as in `slime/rollout/sglang_rollout.py`
https://github.com/THUDM/slime/blob/d78bb43eed5d714d681271fcf8c04f0f28f6b67c/slime/rollout/sglang_rollout.py#L120

### DIFF
--- a/scripts/run-qwen3.5-35B-A3B.sh
+++ b/scripts/run-qwen3.5-35B-A3B.sh
@@ -63,7 +63,7 @@ EVAL_ARGS=(
 )
 
 PERF_ARGS=(
-   --tensor-model-parallel-size 4
+   --tensor-model-parallel-size 2
    --sequence-parallel
    --pipeline-model-parallel-size 1
    --context-parallel-size 1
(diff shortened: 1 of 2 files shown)


Gao016:
hi, i am now also training Qwen3.5-35B-A3B for 100 steps, but from step 0 to step 99, the raw_reward are all 0 and the truncated is 0.9+, very high. 
do you have meet this question? thanks
image

PR #1675, opened 2026-03-07
[Fix][Image] Tmp fix for SGLang new mm processing logic

coding-famer:
Will implement vlm tito soon.

PR #1676, opened 2026-03-07
[Multimodal][Model] Qwen3.5 VL training example/support

coding-famer:
## Notes
- Requires a patched Megatron Bridge (`qwen35` branch) until upstream Megatron&Megatron Bridge is upgraded
- GDN packing not yet supported in Megatron, so `--qkv-format bshd` and `--micro-batch-size 1` are required

Depends on #1648 #1649 #1675


coding-famer:
The curve looks greaaaat
image
image
image

PR #1575, opened 2026-02-11
Support qwen3-next MTP Training

zx3xyy:
# Bug: `linear_qgkv` not handled in weight converter

## Problem

Upstream's `scripts/models/qwen3-next-80B-A3B.sh` enables `--use-gated-attention`, which causes Megatron to create parameters named `linear_qgkv.weight` instead of `linear_qkv.weight`. But the weight converters (`megatron_to_hf/qwen3_next.py` and `mbridge/qwen3_next.py`) only handle `linear_qkv`, causing a crash during weight sync.

## Root Cause

In Megatron's `megatron/core/transformer/attention.py` (line 891):

```python
if self.config.use_gated_attention:
    self.linear_qgkv = build_module(...)   # Q+G+K+V fused
else:
    self.linear_qkv = build_module(...)    # Q+K+V fused
```

MTP layers inherit this config, so they also use `linear_qgkv`.

## Reproduction

Test script that loads the upstream converter and calls it with `linear_qgkv` parameter names (as produced by Megatron when `--use-gated-attention` is set):

```python
import re, torch, traceback
from types import SimpleNamespace

# Load the upstream converter function directly
with open('slime/backends/megatron_utils/megatron_to_hf/qwen3_next.py') as f:
    source = f.read()
exec(compile(source, 'qwen3_next.py', 'exec'))

# Qwen3-Next-80B-A3B model config
args = SimpleNamespace(
    num_attention_heads=16,
    num_query_groups=2,
    kv_channels=256,
    hidden_size=2048,
)
param = torch.randn(9216, 2048)

# Test 1: linear_qkv.weight (handled by upstream - works)
try:
    result = convert_qwen3_next_to_hf(args,
        'module.module.decoder.layers.0.self_attention.linear_qkv.weight', param)
    print(f'[OK] linear_qkv.weight -> {[r[0] for r in result]}')
except Exception as e:
    print(f'[FAIL] linear_qkv.weight: {e}')

# Test 2: linear_qgkv.weight (created when --use-gated-attention is set)
try:
    result = convert_qwen3_next_to_hf(args,
        'module.module.decoder.layers.0.self_attention.linear_qgkv.weight', param)
    print(f'[OK] linear_qgkv.weight -> {[r[0] for r in result]}')
except ValueError as e:
    traceback.print_exc()

# Test 3: linear_qgkv.layer_norm_weight
try:
    result = convert_qwen3_next_to_hf(args,
[...]

PR #1702, opened 2026-03-10
Fix/qwen3 5 mtp bridge

huang3eng:
## Summary

This PR fixes several Qwen3.5 MTP-related issues in the HF/Megatron bridge and improves compatibility across Megatron versions.

In the broken state, Qwen3.5 MTP conversion/training could fail in multiple ways:

1. `tools/convert_hf_to_torch_dist.py` could fail with:
   - `KeyError: 'mtp.layers.0.mlp.experts.gate_up_proj'`
2. Some environments could fail during bridge initialization with:
   - `TypeError: TransformerConfig.__init__() got an unexpected keyword argument 'use_gated_attention'`
3. Even when training launched successfully, early `train/mtp_loss` could be abnormally high (observed around `~12`), indicating incorrect MTP weight/config handling.

This PR addresses those issues for Qwen3.5.

## Background / Related PRs

This fix was also motivated by abnormal Qwen3.5 MTP training behavior: when running `scripts/run-qwen3.5-35B-A3B.sh`, early `train/mtp_loss` was observed around `~12`, which is clearly abnormal.

A related prior discussion exists in the qwen3-next MTP training PR:

- `#1575` Support qwen3-next MTP Training  
  https://github.com/THUDM/slime/pull/1575

This PR is **not** a direct port of `#1575`, and the qwen3-next-specific `linear_qgkv` changes are not directly applicable to Qwen3.5. However, `#1575` was useful as a reference because it highlighted that MTP-specific attention/wrapper handling can silently break training quality even when launch/conversion mostly works.

For Qwen3.5, the relevant issues turned out to be:
- incorrect MTP MoE expert mapping
- incorrect `mtp_block_spec` construction
- wrong `eh_proj.weight` column handling
- unconditional `use_gated_attention` config injection on unsupported Megatron versions

These issues together explain both:
- hard conversion failures, and
- the previously observed abnormal `mtp_loss`

After fixing them and re-converting from the original HF checkpoint, early `train/mtp_loss` returns to a normal range (observed around `0.37 ~ 0.44` instead of `~12`).

### 1. Incorrect MTP MoE expert mapping
Qwen3.5 regular MoE layers use fused expert tensors, but MTP MoE layers use per-expert HF tensors.

The previous bridge code reused the regular MoE mapping for MTP layers, which incorrectly mapped MTP expert weights to nonexistent fused HF names such as:

- `mtp.layers.0.mlp.experts.gate_up_proj`

instead of the actual per-expert HF names such as:

- `mtp.layers.0.mlp.experts.42.gate_proj.weight`
- `mtp.layers.0.mlp.experts.42.up_proj.weight`

### 2. Incorrect `mtp_block_spec` source
`_get_gptmodel_args()` used `self.config` instead of the actual current `transformer_layer_spec` when building `mtp_block_spec`, which could produce an inconsistent MTP block spec.

### 3. Wrong `eh_proj` column handling
`eh_proj.weight` was being column-swapped during bridge conversion and raw Megatron-to-HF export. This corrupted the MTP wrapper projection layout and likely contributed to abnormal MTP loss during training.

### 4. Megatron version compatibility
Some Megatron versions do not accept the `use_gated_attention` kwarg in `TransformerConfig`. Passing it unconditionally caused conversion to fail before loading weights.

### Bridge fixes
- add dedicated MTP MLP mapping for Qwen3.5 per-expert MTP tensors
- route MTP transformer-layer MLP parameters through the MTP-specific mapping
- cache and reuse the current `transformer_layer_spec` when building `mtp_block_spec`
- preserve `eh_proj.weight` column order during HF -> Megatron conversion
- only pass `use_gated_attention=True` when the current `TransformerConfig` supports it

### Raw export fixes
- preserve `eh_proj.weight` column order in the raw Megatron -> HF export path as well

### Tests
Add regression tests covering:
- MTP MoE expert mapping
- dense MTP MLP mapping
- `mtp_block_spec` using the current transformer layer spec
- `eh_proj` column order preservation
- compatibility when `TransformerConfig` supports / does not support `use_gated_attention`

### Unit tests
```bash
python -m pytest tests.test_qwen3_5_mtp_bridge_mapping

### DIFF
--- a/slime_plugins/mbridge/qwen3_5.py
+++ b/slime_plugins/mbridge/qwen3_5.py
@@ -1,3 +1,5 @@
+import inspect
+
 import torch
 from megatron.core.models.gpt.gpt_layer_specs import get_gpt_mtp_block_spec
 
@@ -114,6 +116,22 @@ class Qwen3_5Bridge(Qwen2MoEBridge):
         if hasattr(self.hf_config, "text_config"):
             return self.hf_config.text_config
         return self.hf_config
+
+    def _supports_transformer_config_kwarg(self, kwarg_name: str) -> bool:
+        """Check whether the current TransformerConfig accepts a given kwarg."""
+        transformer_config_class = getattr(self, "TransformerConfigClass", None)
+        if transformer_config_class is None:
+            return True
+
+        dataclass_fields = getattr(transformer_config_class, "__dataclass_fields__", None)
+        if dataclass_fields is not None:
+            return kwarg_name in dataclass_fields
+
+        try:
+            signature = inspect.signature(transformer_config_class)
+        except (TypeError, ValueError):
+            return True
+        return kwarg_name in signature.parameters
 
     def _get_transformer_layer_spec(self, vp_stage=None):
         transformer_layer_spec = super()._get_transformer_layer_spec(vp_stage)
@@ -287,9 +305,11 @@ class Qwen3_5Bridge(Qwen2MoEBridge):
             moe_router_pre_softmax=False,
             qk_layernorm=True,
             attention_output_gate=True,
-            use_gated_attention=True,
             **mtp_args,
         )
+
+        if self._supports_transformer_config_kwarg("use_gated_attention"):
+            base_kwargs["use_gated_attention"] = True
 
         # Handle MoE-specific config
         if hasattr(text_config, "num_experts"):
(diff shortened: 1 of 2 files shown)


huang3eng:
@Zhuohao-Li @zhuzilin Please help review it.🙏🙏

PR #1742, opened 2026-03-19
Support qwen3.5 loss mask for multi-turn SFT

huang3eng:
## Background

`--loss-mask-type` defaults to `qwen`. When Qwen3.5 SFT is launched with the default setting, SFT rollout still goes through the legacy `qwen` loss-mask path, which is incompatible with Qwen3.5 multi-turn chat-template behavior and can fail with:

`jinja2.exceptions.TemplateError: No user query found in messages.`

Using `--loss-mask-type qwen3` avoids the immediate crash, but it still does not match Qwen3.5 masking semantics on multi-turn conversations. For historical assistant turns, Qwen3.5 usually keeps only the final answer, while the current `qwen3` path may reconstruct extra reasoning content and supervise unnecessary thinking tokens. This increases token count and slows down SFT.

## Changes

- add `qwen3_5` as a valid `--loss-mask-type`
- implement a dedicated Qwen3.5 multi-turn loss-mask generator based on the fully rendered conversation
- derive token-level supervision from rendered-text character spans via `offset_mapping`
- validate rendered-text tokenization against `apply_chat_template(..., tokenize=True)` output
- add a defensive check in SFT rollout to ensure `token_ids` and `loss_mask` always have the same length
- add a Qwen3.5 SFT script that explicitly uses `--loss-mask-type qwen3_5`
- add unit tests covering single-turn parity, multi-turn divergence, and tool-call flow behavior

## Why this PR

- fixes the Qwen3.5 SFT failure when the default `qwen` loss-mask path is used
- makes multi-turn masking match Qwen3.5 chat-template behavior
- avoids supervising unnecessary historical reasoning tokens
- reduces wasted training tokens and improves SFT efficiency

## Scope

This PR does not change the global default of `--loss-mask-type`. Instead, it introduces a Qwen3.5-specific option and updates the Qwen3.5 SFT entry script to use it explicitly, which keeps existing Qwen/Qwen3 behavior unchanged.

## Testing

- `python -m pytest tests/utils/test_loss_mask_type_qwen35.py`

### DIFF
--- a/slime/utils/mask_utils.py
+++ b/slime/utils/mask_utils.py
@@ -127,19 +127,14 @@ class MultiTurnLossMaskGenerator:
     def gen_multi_turn_loss_mask_qwen3_5(
         self, messages: list[dict], tools: list[dict] = None
     ) -> tuple[list[int], list[int]]:
-        rendered_text = self.tokenizer.apply_chat_template(
-            messages, tokenize=False, tools=tools, return_dict=False
-        )
-        tokenized = self.tokenizer(
-            rendered_text, add_special_tokens=False, return_offsets_mapping=True
-        )
+        rendered_text = self.tokenizer.apply_chat_template(messages, tokenize=False, tools=tools, return_dict=False)
+        tokenized = self.tokenizer(rendered_text, add_special_tokens=False, return_offsets_mapping=True)
         token_ids = tokenized["input_ids"]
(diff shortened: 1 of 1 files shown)


huang3eng:
@zhuzilin @Zhuohao-Li Please help review it. 💗

PR #1788, opened 2026-03-31
[WIP] fix loss oom

lilei199908:
before optimize
image
after optmize
image
grad norm
image

### DIFF
--- a/slime/backends/megatron_utils/loss.py
+++ b/slime/backends/megatron_utils/loss.py
@@ -412,6 +412,10 @@ def get_log_probs_and_entropy(
         assert max_seq_lens is not None
         logits = logits.view(-1, logits.size(-1))
 
+    # Apply rollout temperature scaling to logits to match rollout-time log-probs.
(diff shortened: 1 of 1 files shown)


slime/utils/ppo_utils.py
```diff
@@ -646,32 +647,38 @@ def chunked_gae(
...
-                entropy = torch.cat(entropys, dim=0)
         else:
-            log_prob = compute_log_probs(logits.clone(), tokens, tp_group)
             if with_entropy:
-                entropy = compute_entropy_from_logits(logits.clone(), tp_group)
+                with entropy_ctx:
+                    entropy_input = logits.clone() if need_entropy_grad else logits
+                    entropy = compute_entropy_from_logits(entropy_input, tp_group)
+
+            log_prob = compute_log_probs(logits.clone(), tokens, tp_group)
```
zhuzilin, line 681: I wonder if some of the code movement (moving log_prob down here) is necessary. And also if we don't need to calculate the grad of entropy, do we still need to do `logits.clone()`?

PR #1842, opened 2026-04-17
Fix double prepare_grads / loss-scaler-double-update in train_one_step

jthomy:
When training with fp16, and therefore args.check_for_nan_in_loss_and_grad=False, train_one_step calls optimizer.prepare_grads() and then optimizer.step(). Megatron's MixedPrecisionOptimizer.step() calls prepare_grads() internally, so prepare_grads runs twice per step causing a doubly unscaled unscaling, causing extremely small gradients for fp16. 

Since optimizer.step() automatically skips the update when inf is detected, we can first run optimizer.step and check for if it was a valid step afterwards. The behavior for non-fp16 is unchanged.

Results with a GLM 4.7 30B Flash:
BF16 (for comparison)
image
FP16 without fix (zero means the grad scaler was too high and is reducing):
image
FP16 with fix:
image

### DIFF
--- a/slime/backends/megatron_utils/model.py
+++ b/slime/backends/megatron_utils/model.py
@@ -431,19 +431,6 @@ def train_one_step(
         forward_only=False,
     )
 
-    valid_step = True
-    grad_norm = float("nan")
-    if not getattr(args, "check_for_nan_in_loss_and_grad", True):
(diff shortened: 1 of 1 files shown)


jthomy:
@zhuzilin if you could review/approve, this would be great.

PR #1856, opened 2026-04-24
refactor/ppo

lilei199908:
future TODO: 
1. maybe not need foward_only for values.
2. curret only surpot same parapall between actor and critic, maybe need more flexible.
3. mayne can considerate muti ciritc
4. more test

curret reward compare verl
up slime， down verl
image

### DIFF
--- a/slime/ray/placement_group.py
+++ b/slime/ray/placement_group.py
@@ -103,13 +103,7 @@ def create_placement_groups(args):
         "rollout": (pg, rollout_pg_reordered_bundle_indices, rollout_pg_reordered_gpu_ids),
     }
 
-    if args.use_critic:
-        critic_num_gpus = args.critic_num_nodes * args.critic_num_gpus_per_node
-        logger.info(f"Creating critic placement group with {critic_num_gpus} GPUs...")
-        c_pg, c_bundle_indices, c_gpu_ids = _create_placement_group(critic_num_gpus)
-        result["critic"] = (c_pg, c_bundle_indices, c_gpu_ids)
-    else:
-        result["critic"] = None
+    result["critic"] = result["actor"] if args.use_critic else None
 
     return result
 
--- a/slime/ray/placement_group.py
+++ b/slime/ray/placement_group.py
@@ -157,16 +157,15 @@ def create_training_models(args, pgs, rollout_manager):
     )
     # TODO how to decide rollout start id when critic is involved? For now we just require user to specify it via args.
     if args.use_critic:
-        all_start_rollout_ids = list(actor_start_rollout_ids)
         critic_start_rollout_ids = ray.get(critic_init_handles)
-        all_start_rollout_ids.extend(critic_start_rollout_ids)
+        all_start_rollout_ids = critic_start_rollout_ids
     else:
         all_start_rollout_ids = actor_start_rollout_ids
 
     assert len(set(all_start_rollout_ids)) == 1
 
     if args.start_rollout_id is None:
-        args.start_rollout_id = actor_start_rollout_ids[0]
+        args.start_rollout_id = all_start_rollout_ids[0]
 
     actor_model.set_rollout_manager(rollout_manager)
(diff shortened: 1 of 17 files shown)


slime/ray/placement_group.py
```diff
@@ -77,47 +77,42 @@ def _create_placement_group(num_gpus):
...
         "rollout": (pg, rollout_pg_reordered_bundle_indices, rollout_pg_reordered_gpu_ids),
     }
 
+    if args.use_critic:
+        critic_num_gpus = args.critic_num_nodes * args.critic_num_gpus_per_node
+        logger.info(f"Creating critic placement group with {critic_num_gpus} GPUs...")
+        c_pg, c_bundle_indices, c_gpu_ids = _create_placement_group(critic_num_gpus)
+        result["critic"] = (c_pg, c_bundle_indices, c_gpu_ids)
+    else:
+        result["critic"] = None
```
zhuzilin:
```suggestion
result["critic"] = result["actor"] if args.use_critic else None
```

slime/utils/arguments.py
```diff
@@ -829,6 +829,19 @@ def add_algo_arguments(parser):
                     "This is useful for sft or custom loss function."
                 ),
             )
+            parser.add_argument(
+                "--custom-advantage-fn",
```
zhuzilin: please follow the naming convention, use `--custom-xxx-function-path`

slime/utils/arguments.py
```diff
@@ -1451,6 +1464,72 @@ def parse_args(add_custom_arguments=None):
...
+                try:
+                    value = type(original)(value)
+                except (ValueError, TypeError):
+                    pass
+        setattr(critic_args, key, value)
+
+    # Critic-specific: disable features that only apply to actors
+    critic_args.kl_coef = 0
+    critic_args.use_opd = False
+    critic_args.normalize_advantages = False
```
zhuzilin: why do we need to turn off normalize adv here?

slime/utils/misc.py
```diff
@@ -94,6 +94,21 @@ def should_run_periodic_action(
     return (step % interval == 0) or (num_rollout_per_epoch is not None and step % num_rollout_per_epoch == 0)
 
 
+def critic_values_by_actor_worker(actor_model, critic_model, per_worker_values):
```
zhuzilin: if we only supports setting critic and actor as the same parallel config, maybe we can simply pass the data from critic to actor and leave this change to future PRs.

train.py
```diff
@@ -79,22 +73,28 @@ def save(rollout_id):
...
-            if rollout_id >= args.num_critic_only_steps and not args.critic_train_only:
-                ray.get(actor_model.async_train(rollout_id, rollout_data_ref))
-            ray.get(critic_train_handle)
+            # Sequential critic -> actor pipeline: critic trains and returns values,
+            # which become the old_values for the actor's advantage computation.
+            value_refs = critic_model.async_train(rollout_id, rollout_data_ref)
+            per_worker_values = ray.get(value_refs)
+
+            if actor_trains_this_step:
+                actor_external = critic_values_by_actor_worker(actor_model, critic_model, per_worker_values)
```
zhuzilin: if we don't need `critic_values_by_actor_worker`, can we pass the value_refs to actor instread of materialize it in the control process?

train_async.py
```diff
@@ -25,12 +25,11 @@ def train(args):
     # create the actor and critic models
     actor_model, critic_model = create_training_models(args, pgs, rollout_manager)
 
-    # always update weight first so that sglang has the loaded weights from training.
-    if not args.critic_train_only:
-        actor_model.update_weights()
+    # Always push actor weights to rollout once weights are loaded.
+    actor_model.update_weights()
```
zhuzilin, line 29: why do we need to do weight updation when `critic_train_only=True`?

zhuzilin (approved):

PR #1913, opened 2026-05-15
fix: use getattr for sglang_speculative_algorithm to avoid AttributeError

none0663:
When sglang args are not parsed (e.g. debug_train_only mode), the args namespace does not have the sglang_speculative_algorithm attribute, causing an AttributeError in _compute_spec_metrics. Using getattr with a default of None safely handles both cases.
Clipboard_Screenshot_1778838539


none0663:
fix in this pr https://github.com/THUDM/slime/commit/799893808039c1d61a1bf369718fe817000ff652

PR #2000, opened 2026-06-01
[docs] fix broken example links in agent.md

ShaoxunZeng:
The `_examples_synced/` paths appear broken when reading the source on GitHub. Switch to `../../../examples/<name>/README.md`.


zhuzilin:
Sorry we cannot change the path here. The path is used within: https://thudm.github.io/slime/get_started/agent.html

Issue #1951, opened 2026-05-27
[Question] retool example: compute_log_probs(logits.clone(), tokens, tp_group) torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 58.15 GiB.


rnwang04:
### Your Question

用https://github.com/THUDM/slime/tree/main/examples/retool 在4卡H200上进行训练，20~50步训练后出现
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 58.15 GiB. GPU 0 has a total capacity of 139.81 GiB of which 47.69 GiB is free. Process 22249 has 3.24 GiB memory in use. Including non-PyTorch memory, this process has 88.83 GiB memory in use. Of the allocated memory 82.55 GiB is allocated by PyTorch, and 2.06 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
                               ^^^^^^^^^^^^^^
    log_prob = compute_log_probs(logits.clone(), tokens, tp_group)
   File "/root/slime/slime/utils/ppo_utils.py", line 672, in calculate_log_probs_and_entropy
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
log_prob, entropy = calculate_log_probs_and_entropy(
```

运行参数和example提供的rl脚本完全一致。
不太明白设置了`--max-tokens-per-gpu 9216`之后，每一步的训练显存依然可能不一样吗？这种训练一定步数后OOM有什么可能的原因呢。

### What I've Tried

- 两次运行这个example都遇到一样的问题

### Environment (if relevant)

- slime version: docker `slimerl/slime:v0.2.4`
- Python version: python3.12
- PyTorch version:
- CUDA/ROCm version:
- GPU type and count: 4 x H200
- OS:

### Additional Context

第一次遇到这个报错的完整log：
```bash
[...]

PR #2011, opened 2026-06-02
perf(ppo): reduce log-prob + entropy cross-entropy peak memory

Mantissagithub:
## What this is

A memory optimization for the log-prob/entropy cross-entropy in `slime/utils/ppo_utils.py`. It:

- fuses the log-prob and entropy computation into a single autograd Function, so it keeps **one** working copy of the logits instead of two clones;
- makes the backward reuse Megatron's saved `softmax` buffer **in place** (no extra full-vocab temps), mirroring `VocabParallelCrossEntropy.calculate_gradients`;
- clones only when autograd would observe Megatron's in-place mutation (`_clone_if_grad_tracked`), keeping the no-grad ref/old-logprob path lean and fixing a `view … modified inplace` crash on the differentiable path;
- wires the existing `--log-probs-chunk-size` into `examples/retool/retool_qwen3_4b_rl.sh`.

## Scope — this is a mitigation, not the whole fix for #1951

I want to be upfront: **this PR does not, on its own, resolve the #1951 OOM.** Digging into that issue (details [there](https://github.com/THUDM/slime/issues/1951)), the failing `58.15 GiB` allocation is a single `[T, 76032]` fp32 tensor with `T ≈ 205,280` — i.e. an `actor_train` micro-batch ~20× larger than `--max-tokens-per-gpu 9216`. That happens because the reporter is on `slimerl/slime:v0.2.4`, which predates the token-balanced training scheduler (`slime/utils/dp_schedule.py`). **The real fix for that OOM is bounding `T`, which current `main` already does.**

So this PR is best viewed as **complementary headroom** for the cross-entropy once `T` is bounded — not the #1951 fix. I've used `Relates to #1951` rather than `Fixes` for that reason. Happy to drop the retool example change if you'd prefer this stay a pure `ppo_utils` optimization.

## Changes
- `slime/utils/ppo_utils.py` — fused `_VocabParallelLogProbsAndEntropy` + in-place backward + `_clone_if_grad_tracked`.
- `examples/retool/retool_qwen3_4b_rl.sh` — set `--log-probs-chunk-size 1024`.
- `tests/test_logprob_entropy_fused.py` (+ CI registration) — forward parity, chunked==unchunked, bf16-tolerance backward, and a TP=2 gloo gradcheck vs a full-vocab reference.
- `tools/repro_1951.py` — synthetic single-rank peak-memory reproducer.
- `slime/backends/megatron_utils/model.py` — opt-in `SLIME_MEM_PROBE` peak logger (env-gated; happy to drop if it's noise).

## Measurements (1×H200, real `megatron-core`, fp32, `--with-entropy --backward`)

Where it helps — fixed `T`, per-card vocab `V=76032`, sweeping `--log-probs-chunk-size`:

| chunk_size | before | this PR |
| --- | --- | --- |
| `-1` (off) | 51.06 GiB | 32.46 GiB |
| `1024` | 38.05 GiB | 27.83 GiB |
| `512` | 37.59 GiB | 27.82 GiB |

And where it doesn't — sweeping the micro-batch token count `T` (chunk `1024`):

| `T` | input `[T,76032]` | before | this PR |
| --- | --- | --- | --- |
| 9,216 | 2.61 GiB | 14.42 GiB | 7.84 GiB |
| 51,200 | 14.50 GiB | 79.82 GiB | 43.56 GiB |
| 103,756 | 29.39 GiB | OOM | 88.24 GiB |
| 205,280 | 58.14 GiB | OOM | OOM |

So it roughly halves the cross-entropy peak in the regime that fits, and turns the ~29 GiB-class step from OOM into a survivable one — but at `T≈205k` it still OOMs, because the logits tensor itself is 58 GiB. No cross-entropy change can help that; bounding `T` is what does.

## Correctness
Backward matches naïve autograd within bf16 tolerance (rel ≲ 0.002) at TP=1 (entropy-active and entropy-zero) and via the TP=2 gloo gradcheck. `pytest tests/test_logprob_entropy_fused.py tests/test_chunked_gae.py` passes; pre-commit clean. The bf16 grad return matches Megatron's fused CE.

## Notes / limitations
- The backward is hand-written; verified at TP=1 and TP=2 only — larger-TP behavior leans on the existing GPU CI suite.
- This reduces the cross-entropy *working set*, not the input logits tensor.
- Follow-up idea (separate PR): filter logits by `loss_mask` before CE to shrink `T` itself.

Relates to #1951.


Mantissagithub:
@huang3eng please help to review this when possible :)

huang3eng:
@Mantissagithub Could we add a small entropy-only backward test? The backward has a separate branch when grad_log_prob is None, and covering it would make the fused autograd behavior easier to trust ～

Mantissagithub:
Yeah sure @huang3eng, thanks for catching this. Added `test_fused_entropy_only_backward_matches_naive_reference`, which backprops through entropy only and checks the grad against a naive reference.

While wiring it up I ran into something: the `grad_log_prob is None` branch didn't seem reachable through autograd, since the Function defaulted to `set_materialize_grads(True)` — so the unused `log_prob` output's grad came in as a full-vocab **zeros** tensor rather than `None`. If I'm reading that right, a plain entropy-only test would pass without ever touching that branch.

So I added `ctx.set_materialize_grads(False)` in `forward` to make it actually reachable. It also avoids allocating that `[seq, vocab]` zeros tensor on single-output backward, which seemed in line with the memory cleanup here. The test now hits the `None` path, and the existing both-grads path looks unchanged — but please double-check my reasoning, happy to adjust if I've missed something.

Full suite (`tests/test_logprob_entropy_fused.py`, CPU):

| Test | Result |
| --- | :---: |
| `test_fused_forward_matches_naive_reference` | ✅ PASSED |
| `test_chunked_matches_unchunked[False]` | ✅ PASSED |
| `test_chunked_matches_unchunked[True]` | ✅ PASSED |
| `test_no_entropy_chunked_backward_preserves_input_grad` | ✅ PASSED |
| `test_fused_backward_matches_naive_reference_with_bf16_tolerance` | ✅ PASSED |
| `test_fused_entropy_only_backward_matches_naive_reference` *(new)* | ✅ PASSED |
| `test_tp2_fused_backward_matches_full_vocab_reference` | ✅ PASSED |

**7 passed**

huang3eng:
LGTM~ @zhuzilin

PR #2017, opened 2026-06-04
feat: add --balance-by-flops for FLOPs-balanced micro-batching

HaoDong0027:
## Summary

This PR adds `--balance-by-flops` for FLOPs-aware micro-batch partitioning in dynamic batching. Instead of balancing by token count alone, it uses `coeff * L + L²` per sample to account for the quadratic cost of attention. The linear coefficient is auto-computed from model config (hidden_size, FFN, SwiGLU, MoE experts/topk/shared).

Changes:

- Add `--balance-by-flops` flag and auto-compute `workload_coeff` in `arguments.py`.
- Replace first-fit token packing with Karmarkar-Karp FLOPs-balanced partitioning for micro-batch grouping and DP rank assignment in `dp_schedule.py`.
- Add `calculate_workload()` in `seqlen_balancing.py`.

## Motivation

Dynamic batching packs micro-batches by token count (`Σ L`), but attention FLOPs scale as `O(L²)`. When sequence lengths vary widely, a rank with a few long sequences has much higher compute cost than one with many short sequences. Since all ranks synchronize at all-reduce, the slowest rank determines step time.

`--balance-by-flops` uses `coeff * L + L²` as the balancing metric. This operates purely at the micro-batch partitioning level and does not affect downstream context-parallel splitting, so it composes directly with CP.

## Experiments

All experiments compare `--balance-by-flops` (with `--balance-data`) against a baseline using only `--balance-data` (token-sum KK balancing). Both use `--use-dynamic-batch-size`. Statistics exclude warmup steps.

### Setup

| | SFT Dense | SFT MoE | RL Dense | RL MoE |
| --- | --- | --- | --- | --- |
| **GPU** | 4×L20Z | 8×L20Z | 8×L20Z | 8×L20Z |
| **Model** | Qwen3-4B | Qwen3.5-35B-A3B | Qwen3-4B | Qwen3-30B-A3B |
| **Type** | Dense | MoE | Dense | MoE |
| **Task** | SFT | SFT | RL (GRPO) | RL (GRPO) |
| **TP / PP / EP / DP** | 1 / 1 / — / 4 | 2 / 1 / 8 / 4 | 2 / 1 / — / 4 | 4 / 1 / 8 / 2 |
| **max_tokens_per_gpu** | 9,216 | 8,192 | 9,216 | 20,480 |
| **global_batch_size** | 128 | 128 | 256 | 256 |
| **max_response_len** | — | — | 8,192 | 8,192 |

### Results

| | SFT Dense | SFT MoE | RL Dense | RL MoE |
| --- | ---: | ---: | ---: | ---: |
| **actor_train_time Δ** | −12.7% | −30.1% | −14.0% | −20.6% |
| **actor_train_tflops Δ** | +13.7% | +23.4% | +4.5% | +23.7% |
| **actor_train_tok/s Δ** | +13.7% | +23.4% | +6.0% | +24.0% |

## Notes

- `--balance-by-flops` requires `--use-dynamic-batch-size`.
- The flag is orthogonal to `--balance-data`; when both are set, FLOPs-based weights are used for both micro-batch packing and DP rank assignment.

### DIFF
--- a/slime/utils/arguments.py
+++ b/slime/utils/arguments.py
@@ -1786,23 +1786,6 @@ def slime_validate_args(args):
     if getattr(args, "balance_by_flops", False):
         assert args.use_dynamic_batch_size, "--balance-by-flops requires --use-dynamic-batch-size"
 
-        h = getattr(args, "hidden_size", None)
-        ffn = getattr(args, "ffn_hidden_size", None)
-        if h is None or ffn is None:
-            args.workload_coeff = 24576
-        else:
-            num_experts = getattr(args, "num_experts", None)
-            if num_experts is not None and num_experts > 1:
-                topk = getattr(args, "moe_router_topk", 1)
-                moe_ffn = getattr(args, "moe_ffn_hidden_size", ffn)
-                shared = getattr(args, "moe_shared_expert_intermediate_size", None) or 0
-                d_ff = moe_ffn * topk + shared
-            else:
-                d_ff = ffn
-            ffn_mul = 3 if getattr(args, "swiglu", False) else 2
-            args.workload_coeff = 2 * h + ffn_mul * d_ff // 2
-        logger.info(f"balance-by-flops: workload_coeff={args.workload_coeff}")
-
     if args.eps_clip_high is None:
         args.eps_clip_high = args.eps_clip
 
--- a/slime/utils/arguments.py
+++ b/slime/utils/arguments.py
@@ -687,17 +687,6 @@ def get_slime_extra_args_provider(add_custom_arguments=None):
             )
 
             parser.add_argument(
-                "--exact-flops-balance",
-                action="store_true",
-                default=False,
-                help=(
-                    "When used with --balance-by-flops, compute exact per-sample FLOPs via "
-                    "calculate_fwd_flops instead of the lightweight coeff*L + L² approximation. "
-                    "Useful for ablation studies comparing approximate vs exact FLOPs balancing."
-                ),
-            )
-
-            parser.add_argument(
                 "--use-dynamic-batch-size",
                 action="store_true",
                 default=False,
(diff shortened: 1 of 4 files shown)


HaoDong0027:
@huang3eng Please help to review.

huang3eng:
LGTM~ @zhuzilin

slime/utils/arguments.py
```diff
@@ -1769,6 +1783,26 @@ def slime_validate_args(args):
...
+            num_experts = getattr(args, "num_experts", None)
+            if num_experts is not None and num_experts > 1:
+                topk = getattr(args, "moe_router_topk", 1)
+                moe_ffn = getattr(args, "moe_ffn_hidden_size", ffn)
+                shared = getattr(args, "moe_shared_expert_intermediate_size", None) or 0
+                d_ff = moe_ffn * topk + shared
+            else:
+                d_ff = ffn
+            ffn_mul = 3 if getattr(args, "swiglu", False) else 2
+            args.workload_coeff = 2 * h + ffn_mul * d_ff // 2
```
zhuzilin: can we reuse the `calculate_fwd_flops` function here?

HaoDong0027: Maybe feasible，I'll test it on Qwen3-30B-A3B and report back.

HaoDong0027: We ran an ablation on **Qwen3-30B-A3B** (RL GRPO, TP=4, EP=8, DP=2) comparing baseline (token-sum KK), lightweight approximation (`coeff*L + L²`), and exact FLOPs (`calculate_fwd_flops`):

| Metric | Baseline (token-sum) | Approx (`coeff*L+L²`) | Exact (`calculate_fwd_flops`) |
| --- | ---: | ---: | ---: |
| **actor_train_time** | 119.48s | 94.80s (**-20.7%**) | 94.11s (**-21.2%**) |
| **actor_train_tflops** | 49.58 | 61.26 (**+23.5%**) | 59.79 (**+20.6%**) |
| **actor_train_tok/s** | 14,535 | 18,006 (**+23.9%**) | 18,633 (**+28.2%**) |

Since the approximate and exact methods provide nearly identical speedups over the baseline, we refactored the implementation to reuse calculate_fwd_flops directly rather than introducing an additional workload estimation function. Please see the latest commit for details.

liujia-cc:
@HaoDong0027 Good Job👍! I have a small question, this change will reduce the number of micro-batches, and you use kk for packing, would that affect the time of the experiment, the origin algorithm is first fit.
image

zhuzilin (approved):

zhuzilin:
Great work! I think this is a better default behavior for slime.

Instead of keeping --balance-by-flops as a separate opt-in mode, I think we can directly make it the default implementation behind --balance-data, since the old token-count-based --balance-data is suboptimal when sequence lengths vary significantly.

I’ll make this change and use your implementation as the default balancing strategy.

PR #2024, opened 2026-06-05
Log progress while waiting for placement group

luccabb:
## Problem
`_create_placement_group` waits for the placement group with a bare `ray.get(pg.ready())`. When the placement group can't be placed yet — a node's GPUs haven't registered with the Ray GCS, or an autoscaler is still bringing nodes up — this blocks with **no output**, the last log line being `Creating placement group with N GPUs...`. The job looks hung with no indication of whether it's stuck, slow, or scaling. This may underlie some of the "training hangs at startup" reports.

## Fix
Poll `pg.ready()` and log the GPU counts every 30s while waiting:
```
Waiting for placement group of 16 GPUs (elapsed 30s): 8 GPUs registered with Ray, 8 available.
```
The wait stays **unbounded** — no timeout, no behavior change — so autoscaling clusters (where a pending placement group is the signal that triggers scale-up, and a cold start can legitimately take many minutes) are unaffected.

### DIFF
--- a/slime/ray/placement_group.py
+++ b/slime/ray/placement_group.py
@@ -47,7 +47,24 @@ def _create_placement_group(num_gpus):
     pg = placement_group(bundles, strategy="PACK")
     num_bundles = len(bundles)
 
-    ray.get(pg.ready())
+    # Wait for the placement group to be scheduled. Poll rather than a bare
+    # ray.get(pg.ready()) so the wait is observable: when it can't be placed yet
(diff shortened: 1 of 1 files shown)


luccabb:
wondering if its a good idea to timeout?

zhuzilin (approved):

zhuzilin:
thanks a lot for this! yeah the origin logging was not intuitive.

> wondering if its a good idea to timeout?

hmm... I've not sure about this...

PR #2034, opened 2026-06-08
[docs] Add AgentCore RL Toolkit to ecosystem list

lyzustc:
Add [AgentCore RL Toolkit](https://github.com/awslabs/agentcore-rl-toolkit) to the list of ecosystem built on slime in README.


Teq2412:
@zhuzilin This is the ART package we mentioned, thanks in advance for reviewing effort !

PR #2036, opened 2026-06-09
fix(search-r1): stop generation at </search> and </answer>

yichuan-w:
### Problem
In `examples/search-r1/generate_with_search.py`, the multi-turn rollout never tells the inference engine to stop at the tool/answer boundary. So the model keeps generating after `</search>` / `</answer>` (junk, sometimes fabricating new `Question:`s).

The example only trims that via `postprocess_responses` (string split), which is **disabled when `return_logprob=True`** (needed to keep token/logp aligned for TIS). So with logprob collection on, the trailing junk:
- stays in the trajectory and is trained on (`loss_mask=1`), and
- breaks `is_valid_sequence` (content after `</answer>` → format invalid → lower reward).

### Fix
Pass `stop=["</search>", "</answer>"]` in `sampling_params`. slime already sets `no_stop_trim=True`, so the closing tag is kept and token/logp stay aligned natively — no string post-processing needed, and it works whether `return_logprob` is on or off.

### DIFF
--- a/examples/search-r1/generate_with_search.py
+++ b/examples/search-r1/generate_with_search.py
@@ -157,6 +157,21 @@ async def generate(args, sample: Sample, sampling_params) -> Sample:
     loss_mask = []
     rollout_log_probs = [] if SEARCH_R1_CONFIGS["return_logprob"] else None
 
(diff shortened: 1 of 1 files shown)


yichuan-w:
image
compare two curve, the fix one is much better

PR #2059, opened 2026-06-12
fix(rollout): retry transient Ray ActorUnavailableError during engine bringup

EazyReal:
## What changed

During colocated rollout-engine startup the SGLang engine actors can already be healthy while Ray briefly reports a peer actor as unavailable due to a transient control-plane heartbeat / gRPC `UNAVAILABLE` miss, surfacing as `ray.exceptions.ActorUnavailableError`. Both bringup waits — `RolloutManager.__init__` awaiting the deferred rollout init handles, and the encoder-disaggregated startup awaiting encoder init handles — were bare `ray.get(...)` calls that aborted startup on that transient error even though re-awaiting the same already-submitted ObjectRefs would succeed.

- Add `slime/utils/retry.py`: `retry_with_backoff(thunk, *, should_retry, what, ...)`, a bounded synchronous retry with exponential jittered backoff. The helper owns retry count, backoff, jitter, logging, and exhaustion; the caller owns the retry predicate. Backoff/jitter mirror the existing async idiom in `slime/rollout/rm_hub/__init__.py`.
- In `slime/ray/rollout.py`, wrap the two bringup waits with the helper, retrying only `ActorUnavailableError`. The same ObjectRefs are re-awaited — no engine actors are recreated. Permanent failures (CUDA OOM, config/assert errors, `ActorDiedError`) are not `ActorUnavailableError` and propagate immediately.

## Why

A momentary heartbeat miss at the bringup wait boundary is recoverable: the engine is up and re-driving `ray.get` on the same refs returns. Aborting a multi-node rollout startup over a transient control-plane blip wastes a full bringup. Scope is deliberately the initial bringup waits only; engine recovery (`ServerGroup.recover`) is unchanged.

## Validation

- `PYTHONPATH=. pytest tests/test_retry.py` — helper recovers after transient failures, exhaustion re-raises the last error, predicate-rejected errors propagate immediately with no retry, and `KeyboardInterrupt`/`SystemExit` are never intercepted. Registered in the cpu-unittest list in `.github/workflows/pr-test.yml.j2`.
- Call-site reasoning: both `_wait_rollout_engine_bringup_handles` uses pass the original handles list unchanged into `ray.get`, so a retry re-awaits the same refs rather than recreating engines.


EazyReal:
@zhuzilin could you review this one? It retries the narrow Ray ActorUnavailableError window during engine bringup, which is transient in Ray, without broadening retry behavior around generation or real rollout failures.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This retries transient Ray ActorUnavailableError during rollout engine bringup, reducing flaky startup failures. Could you review when you have bandwidth?

PR #2061, opened 2026-06-12
fix(rollout): apply sample filter after every train rollout function

EazyReal:
## What changed

`--rollout-sample-filter-path` now runs in `call_rollout_fn` after every train `--rollout-function-path` returns, including custom functions. The in-`generate_rollout_async` call is removed so the default path is not applied twice. The rollout manager is unchanged.

## Why

The filter used to live only inside the default sglang rollout function. A custom `--rollout-function-path` that did not copy that hook skipped filtering. Wrapping at `call_rollout_fn` is the current slime function-based equivalent of applying the hook on the default class rollout.

## Validation

```bash
pytest tests/test_rollout_sample_filter.py
```

Registered in the `cpu-unittest` matrix (`pr-test.yml.j2` after `test_filter_long_prompt.py`).


EazyReal:
@zhuzilin could you review this one? It moves rollout sample filtering to the rollout manager boundary so custom and default rollout functions share the same filtering contract.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This applies rollout sample filtering in the rollout manager so filtered samples are handled before training-data conversion. Could you review when you have bandwidth?

PR #2062, opened 2026-06-12
fix(grpo): group-relative baseline by group_index and rollout_id

EazyReal:
## What changed

`_post_process_rewards` now calls `normalize_rewards_by_rollout`. Prompt groups use `Sample.group_index` when every sample has one; otherwise the old positional `n_samples_per_prompt` layout. Within a prompt, `rollout_id` (else `index`) is one rollout: siblings must share one full reward, the baseline counts that rollout once, and the advantage is broadcast.

Does not reshape by `n_samples_per_prompt`, and does not sum sibling rewards (`get_trajectory` already assigns the full trajectory reward to every segment).

## Why

Uneven or fanned rollouts made the old reshape fall through to a single global-mean baseline, mixing prompts and counting fanned segments more than once.

## Validation

```bash
python tests/test_normalize_rewards_by_rollout.py
```

Registered in the `cpu-unittest` matrix (`pr-test.yml.j2` after `test_advantage_whiten_cp.py`).


EazyReal:
@zhuzilin could you review this one? GRPO baselines need to be computed per prompt when rollouts are fanned or uneven; this avoids mixing rewards across prompts and biasing advantages.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This applies a per-prompt baseline for fanned or uneven rollouts, which keeps GRPO normalization correct with non-uniform samples. Could you review when you have bandwidth?

PR #2005, opened 2026-06-02
[coding-agent-rl] Refactor coding-agent RL: turn-node TrajectoryManager + pluggable harness layer

jingshenghang:
## What

Refactors the coding-agent RL rollout subsystem (`slime/agent/`, `examples/coding_agent_rl/`)
around two structural changes, plus a full test-suite reorganization. Net diff is
mostly a rewrite, not new surface area (~4.9k +/3.3k − across 28 files).

### 1. Turn-node `TrajectoryManager` rewrite of `trajectory.py`

`slime/agent/trajectory.py` is rewritten in place. The old implementation
linearized a rollout into reward-split *segments*; the new `TrajectoryManager`
models a session as a per-sid **message tree** of turn nodes:

- `record_turn(TurnRecord)` feeds each turn (prompt messages + the served model's
  sglang snapshot) into the tree.
- `get_trajectory()` linearizes the tree into a `list[Sample]` of loss-masked
  training rows.
- Tolerates TITO re-tokenization drift via fork/replace on the common token prefix
  (strict exact-prefix linearization; drifted response spans are masked).
- `TurnRecord` is the explicit adapter↔manager contract (prompt/output ids,
  finish_reason, log-probs).

### 2. Pluggable harness layer

Coding-agent CLIs are now swappable behind `slime/agent/harness/`:
`BaseHarness` + `HarnessContext`, with `ClaudeCodeHarness` and `CodexHarness`
implementations. Adapters (`anthropic.py`, `openai.py`) are de-scaffolded onto a
shared `common.py` pipeline; harnesses are `SingletonMeta`-backed (no module-level
singletons).

### 3. Example + env cleanup

- `examples/coding_agent_rl/sandbox.py` → `swe.py` (SWE eval split out);
  sandbox provisioning consolidated into `slime/agent/sandbox.py`.
- Env vars unified under `SLIME_AGENT_*` / `ADAPTER_*`.
- Fan-out / sandbox / reply-path notes moved from the launcher script into
  `README.md`; launcher trimmed.

### 4. Test reorganization

`tests/test_agent_{adapters,sdk_adapters,trajectory}.py` removed in favor of a
`tests/test_agent/` package: `test_trajectory_manager_branching.py`,
`test_adapters.py`, `test_harness.py`, `test_agent_rollout_cpu.py` (plus shared
`_fakes.py` / `_dump_helpers.py`). New `agent-test` job in the CI matrix
(`pr-test.yml.j2`) runs them on every push/PR.

PR #2063, opened 2026-06-12
fix(agent): render OpenAI tool-call arguments as a mapping for chat templates

EazyReal:
## Problem

In the OpenAI HTTP adapter (`slime/agent/adapters/openai.py`), `_normalize_tool_call` builds the assistant message that `_translate_chat_messages` collects into `chain.chat_messages`, which `render_token_ids` then feeds to `tokenizer.apply_chat_template`. That message is the *render boundary*. It serialized `tool_calls[].function.arguments` to a JSON **string** via `json_arguments`. Qwen-family chat templates iterate `tool_call.arguments.items()`, which expects a mapping — given a string they either iterate it character-by-character or raise a template error, so the rendered prompt diverges from the tokens the policy actually generated. In a token-capturing rollout that is a **rollout/train token desync**: the policy is trained on tokens it never sampled.

## Before vs After

Same input — an assistant turn whose tool call carries `arguments` as the JSON string echoed on the OpenAI wire:

```python
openai._normalize_tool_call(
    {"function": {"name": "lookup", "arguments": '{"q": "slime"}'}}
)
```

**Before** — `arguments` stays a JSON string, so the rendered assistant message is:

```python
{"type": "function",
 "function": {"name": "lookup", "arguments": '{"q": "slime"}'}}   # str
```

The Qwen template's `arguments.items()` then runs on a string → `AttributeError: 'str' object has no attribute 'items'` (or, on templates that tolerate it, per-character iteration that emits a garbled tool block). Either way the rendered prompt no longer matches the sampled tokens.

**After** — the wire string is decoded back to the mapping the template expects:

```python
{"type": "function",
 "function": {"name": "lookup", "arguments": {"q": "slime"}}}     # dict
```

`arguments.items()` now yields `[("q", "slime")]` and the tool call renders correctly.

Hostile inputs that are valid JSON but not a mapping are wrapped so `.items()` can never crash:

```python
openai._normalize_tool_call(
    {"function": {"name": "lookup", "arguments": "[1, 2]"}}
)
# arguments -> {"_raw_arguments": [1, 2]}
```

Falsy non-dict values are real argument payloads, not "no arguments", and are preserved the same way — only `None` (or an empty wire string) maps to `{}`:

```python
dict_arguments(0)     # -> {"_raw_arguments": 0}
dict_arguments([])    # -> {"_raw_arguments": []}
dict_arguments(None)  # -> {}
```

## Fix

Add a small pure helper `dict_arguments(value) -> dict` in `slime/agent/adapters/common.py` that decodes echoed wire strings to a mapping at the render boundary:

- `dict` passes through unchanged;
- `str` is `json.loads`-decoded (an empty wire string → `{}`: the OpenAI wire encodes "no arguments" as an empty / `"{}"` string);
- `None` → `{}` (no arguments); **every other** non-dict outcome — including falsy values like `0`, `False`, and `[]` — funnels through a `{"_raw_arguments": ...}` sentinel (already a convention in `slime/agent/parsing.py`), so the `-> dict` contract holds for **all** inputs, `.items()` can never crash, and no argument payload is ever silently dropped.

Switch the single render-boundary call site in `_normalize_tool_call` from `json_arguments` to `dict_arguments`.

## Why this is the right fix

- **Default-path safe.** The dict-passthrough and empty/`None` → `{}` branches mean callers already passing a mapping (or no arguments) get bit-identical output; only echoed JSON-string arguments change, and they now match what the chat template expects.
- **Lossless.** Truthiness is the wrong gate for "no arguments": `0`, `False`, and `[]` are payloads the model emitted. Gating the sentinel on `is None` instead of truthiness means the normalization never rewrites a real payload to an empty call.
- **Outbound wire contract preserved.** `_openai_tool_calls` (the response path) is untouched and still emits `function.arguments` as a JSON string, exactly as the OpenAI spec requires. Only the internal render boundary changed.
- **No new abstraction.** It is one pure function reusing the existing `{"_raw_arguments": ...}` sentinel from `parsing.py`; the `-> dict` return type guarantees `.items()` is always safe.
- **CI-verifiable.** `tests/test_agent_adapters.py::test_openai_render_tool_call_arguments_are_dicts` (marked `unit`) asserts the rendered arguments are the decoded mapping `{"q": "slime"}`, that `"[1, 2]"` wraps to `{"_raw_arguments": [1, 2]}`, and that falsy non-dict values are preserved losslessly (`0` → `{"_raw_arguments": 0}`, `[]` → `{"_raw_arguments": []}`, `None` → `{}`). The file already runs in the CPU `agent-adapter-test` job (`num_gpus: 0`), so the test executes with no workflow change.

### DIFF
--- a/slime/agent/adapters/common.py
+++ b/slime/agent/adapters/common.py
@@ -95,6 +95,28 @@ def json_arguments(value: Any) -> str:
     return json.dumps(value, ensure_ascii=False)
 
 
+def dict_arguments(value: Any) -> dict:
+    """Tool-call arguments for chat-template rendering.
+
+    OpenAI wire responses carry ``function.arguments`` as a JSON string; Qwen-family
+    chat templates expect a mapping and iterate ``arguments.items()``. Decode echoed
+    wire strings back to dicts at the render boundary.
+    """
+    if isinstance(value, dict):
+        return value
+    if isinstance(value, str):
+        try:
+            decoded = json.loads(value or "{}")
+        except json.JSONDecodeError:
+            decoded = value
+        if isinstance(decoded, dict):
+            return decoded
+        return {"_raw_arguments": decoded}
+    if value is None:
+        return {}
+    return {"_raw_arguments": value}
+
+
 def render_token_ids(chain: AdapterChain, tokenizer) -> list[int]:
     enc = tokenizer.apply_chat_template(
         chain.chat_messages,
--- a/slime/agent/adapters/openai.py
+++ b/slime/agent/adapters/openai.py
@@ -23,6 +23,7 @@ from aiohttp import web
 from slime.agent.adapters.common import ADAPTER_KEY, REASONING_PARSER_KEY, TOKENIZER_KEY, TOOL_PARSER_KEY
 from slime.agent.adapters.common import AdapterChain as Chain
 from slime.agent.adapters.common import BaseAdapter, call_sglang_generate
+from slime.agent.adapters.common import dict_arguments as _dict_arguments
 from slime.agent.adapters.common import json_arguments as _json_arguments
 from slime.agent.adapters.common import ok_response, render_token_ids, request_session_id
 from slime.agent.adapters.common import stable_hash as _hash
@@ -105,7 +106,7 @@ def _normalize_tool_call(call: dict[str, Any]) -> dict[str, Any]:
         "type": "function",
         "function": {
             "name": name,
-            "arguments": _json_arguments(arguments),
+            "arguments": _dict_arguments(arguments),
         },
     }
     if call.get("id"):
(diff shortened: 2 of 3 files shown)


jingshenghang:
Hi, our refactoring of the agent framework has been merged into the main branch. The Codex/OpenAI testing and validation were insufficient; we welcome pull requests based on our latest refactor.

https://github.com/THUDM/slime/pull/2005

EazyReal:
Closing as superseded by current `main`. The OpenAI tool-call `arguments` are now coerced to a mapping at the chat-template render boundary by `openai._arguments_as_dict()` (called from `_translate_messages`), and `tests/test_agent/test_adapters.py::test_openai_translation_developer_to_system_and_tool_calls_to_dict` already asserts this exact property. The functions this PR edited (`_normalize_tool_call`, `json_arguments`) no longer exist after the adapter refactor (#1979/#1981/#2005/#2124), so there's nothing left to rebase. Thanks!

PR #2064, opened 2026-06-12
fix(metrics): estimate pass@k over complete group_index groups

EazyReal:
## What changed

`compute_pass_rate` keeps its rectangular contract (including the size assert). New `compute_pass_rate_by_group_index` buckets by `group_index` and keeps only groups whose size is exactly `n_samples_per_prompt` / `n_samples_per_eval_prompt`, then calls `compute_pass_rate` on that packed rectangle.

The default converter packs optional `group_indices` (shipped whole like `raw_reward`). Missing/`None` `group_index` keeps the legacy reshape. Incomplete or oversampled groups are dropped from pass@k, not estimated ragged.

## Why

Over-sampled or filtered batches no longer divide `group_size`, so the rectangle assert crashed metric logging. Complete-group pass@k matches Miles identity grouping without changing `compute_pass_rate`'s public signature.

## Validation

```bash
pytest tests/utils/test_metric_utils.py
```

Registered in the `cpu-unittest` matrix (`pr-test.yml.j2` after `test_metric_report.py`).


EazyReal:
@zhuzilin could you review this one? compute_pass_rate assumed rectangular over-sampled batches; this makes the metric ragged-safe so uneven prompt fanout does not produce incorrect pass-rate accounting.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This makes compute_pass_rate handle ragged over-sampled batches instead of assuming every prompt has the same fan-out. Could you review when you have bandwidth?

PR #2066, opened 2026-06-12
[megatron] don't re-assert no_sync_func every step with overlap_grad_reduce

### LINKED ISSUE #1779: [Bug] `assert config.no_sync_func is None` always gets hit second step
### Bug Description

https://github.com/THUDM/slime/blob/5d7ca976039e738ebc2ed003a6f9312b00820810/slime/backends/megatron_utils/model.py#L524-L528

The first step it sets `config.no_sync_func` then the second step it crashes at the assert

### Steps to Reproduce

Run any training with `--overlap-grad-reduce`

### Expected Behavior

`--overlap-grad-reduce` works

### Actual Behavior

Crashes 2nd step

### Environment

any

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

HaozheZhang6:
## Summary

`train()` in `slime/backends/megatron_utils/model.py` sets up `config.no_sync_func` on **every** step, but `config = get_model_config(model[0])` is the model config and persists across steps. With `--overlap-grad-reduce`, the first step passes `assert config.no_sync_func is None` and sets it; the **second step** then trips that same assert and crashes.

```
assert config.no_sync_func is None, (...)   # step 1: None -> ok, sets it
                                            # step 2: not None -> AssertionError
```

## Fix

Guard the setup with `if config.no_sync_func is None:` so the sync funcs are set once. They're constant (the model chunks' own `no_sync` / `start_grad_sync`), so skipping on later steps is a no-op — the only thing the per-step re-run did was trip the assert.

## Reproduction

Run any training with `--overlap-grad-reduce` (from #1779) — it crashes on the 2nd step without this change.

## Notes

- I couldn't add an automated test: reproducing needs a multi-GPU Megatron training run with `--overlap-grad-reduce`, which I can't stand up locally. The fix is small and verified by inspection against the reported repro; happy to add a test if you can point me at a lightweight harness for this path.
- Opening as a **draft** to confirm the approach: the original `assert` also doubled as a guard against a *pre-supplied custom* `no_sync_func`. Under the `is None` guard it becomes redundant, and a hypothetical pre-set custom func would now be left as-is rather than rejected. If you'd rather keep that rejection explicit (e.g. assert only on first setup), I'll adjust.

Fixes #1779


HaozheZhang6:
Added a unit test for the idempotency (`tests/test_overlap_grad_reduce.py`). Pulled the setup into a small `grad_reduce.py` helper so it's testable without a full training step -- it stubs megatron's DDP the same way the other CPU tests do. Also ran it on a single GPU with a real megatron DDP: the 2nd step trips the assert without the guard, not with it.

PR #2068, opened 2026-06-12
fix(sglang): authenticate engine control-plane and router calls

EazyReal:
## What changed

Thread bearer `Authorization` headers through the SGLang HTTP calls slime owns when `--sglang-api-key` / router `--api-key` are configured.

- `bearer_auth_headers` in `slime/utils/http_utils.py` returns no `Authorization` header when the key is unset, so unauthenticated deployments never send `Bearer None`.
- `SGLangEngine` stores the resolved worker key as `self.server_api_key` and uses it for engine control-plane calls: health wait, cache flush, weight update, memory release/resume, pause/continue generation, profile start/stop, and server-info discovery.
- Router auth is applied to worker registration/listing/removal and to default and streaming `/generate` calls. The `/workers` registration payload still forwards the worker key so the router can authenticate traffic it proxies.
- `server_control.abort_servers_until_idle` and `sglang_rollout.abort` authenticate the `/abort_request` + `/v1/loads` loop and the router worker-listing call.
- `flush_cache` fails fast on 401/403 instead of retrying for 60s.

## Why

slime could not drive an SGLang server/router with API-key auth enabled. The post-launch health wait sent `Authorization: Bearer None` when no key was set, and most later control calls were unauthenticated. Router management and generate/abort could be rejected as soon as training tried to register workers, flush cache, sync weights, abort, or generate.

## Validation

CPU unit tests (registered in `.github/workflows/pr-test.yml.j2`):

```bash
python tests/test_sglang_control_plane_auth.py
python tests/test_external_sglang_engines.py
```

`test_sglang_control_plane_auth.py` pins that falsy keys never produce `Bearer None` and that abort-loop HTTP calls carry the worker bearer. `test_external_sglang_engines.py` covers `get_server_info` / discover API-key threading.


EazyReal:
@zhuzilin could you review this one? Engine init/update calls had auth headers, but default and streaming generation still hit the router without them; this centralizes router request headers and closes the remaining authenticated-control-plane gap.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This authenticates SGLang engine and router HTTP calls, including the upstream rank-0 control-plane guard. Could you review when you have bandwidth?

PR #2069, opened 2026-06-12
fix(agent): pool the SGLang client across turns and retry once on pre-connect errors

EazyReal:
## What changed

`call_sglang_generate` no longer opens a fresh `aiohttp.ClientSession` per `/generate` turn, nor a separate short-lived session for `/abort_request`. Each adapter app now owns one pooled `ClientSession` via `cleanup_ctx` (`TCPConnector(limit=0, ttl_dns_cache=300, keepalive_timeout=60)`), so OpenAI and Anthropic turns reuse warm keep-alive connections. Abort requests use the same pooled client with a per-call short timeout.

A `/generate` POST that fails with `aiohttp.ClientConnectorError` is retried exactly once with the same `rid`. Failures that may have reached SGLang, such as disconnects and timeouts, are not retried and still abort by `rid`. HTTP `>= 400` responses preserve upstream behavior: log the response body and raise without retrying.

## Why

Concurrent multi-turn agent rollouts were opening and tearing down a connection per turn, which churned sockets and could raise `ClientConnectorError` before the request reached SGLang. Pooling removes the per-call connection churn. The retry is confined to pre-connect connector failures, so reusing the same `rid` cannot double-generate; anything that may have reached the server is never re-issued.

## Validation

CPU unit tests in `tests/test_agent/test_adapters.py` (already registered in `cpu-unittest`) cover real adapter loopback behavior plus the pooled-client invariants: two turns use the same app-owned client, a connector error retries exactly once with the same `rid`, `ServerDisconnectedError` is not retried and aborts by `rid`, and two connector errors propagate after one retry then abort.


EazyReal:
@zhuzilin could you review this one? It pools the SGLang client across agent turns and retries only the narrow transient pre-connect failure, reducing connection churn without masking persistent rollout errors.

EazyReal:
Gentle follow-up after rebasing and trimming the test/comment noise: this reuses one SGLang client across agent turns and retries only pre-connect failures. That reduces connection churn without retrying requests that may have reached SGLang.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This pools the SGLang client across agent turns and retries transient pre-connect failures, reducing avoidable rollout failures. Could you review when you have bandwidth?

EazyReal:
I have moved this PR back to draft while clarifying the intended retry boundary.

The pooled per-app client is independently useful: it removes per-turn connection churn and gives the client an explicit app shutdown owner. The retry portion is deliberately narrow: retry one connector-stage failure with the same rid, and abort rather than resend after an ambiguous client error.

One deployment-dependent gap remains outside this diff: if sglang_url points to SGLang Router, the router may perform its own retries on retryable HTTP statuses. Could maintainers clarify the expected ownership here?

1. Should the adapter fail the turn after a may-have-reached error and let a higher-level rollout coordinator decide whether to redo it?
2. Or is the adapter itself expected to prioritize availability and resubmit the turn?
3. In normal deployments, does this URL point directly to an engine or through a retrying router?

I propose keeping this PR scoped to connection pooling and the connector-stage retry until that contract is clear, rather than embedding a broader training retry policy in the adapter.

PR #2084, opened 2026-06-15
feat(rl): add off-policy IS correction hook (current policy vs rollout)

EazyReal:
## What changed

- `policy_loss_function` now passes current `pi_theta` log-probs as `cur_log_probs` into the TIS-correction kwargs, alongside frozen `train_log_probs` and `rollout_log_probs`. Existing `vanilla_tis_function` / `icepop_function` ignore the new key via `**kwargs`.
- New `off_policy_is_function` in `slime/utils/ppo_utils.py`: truncated IS with detached weight `clip(pi_theta / pi_rollout)` using `--tis-clip-low` / `--tis-clip` (same bounds as vanilla TIS / icepop). Metric keys `tis` / `tis_clipfrac` / `tis_abs`. `loss_masks` are unchanged.
- Opt-in via the existing hook: `--use-tis --custom-tis-function-path slime.utils.ppo_utils.off_policy_is_function`.
- Independent of named CISPO (`--advantage-estimator cispo` is unchanged). On a REINFORCE base this is the compositional CISPO-style correction (current policy vs rollout), not a second named estimator.

## Why

Vanilla TIS uses frozen `pi_theta_old / pi_rollout`, which matches current-policy IS only in the single-update-per-rollout limit. Passing `cur_log_probs` lets a TIS hook correct both train/inference mismatch and async staleness.

## Validation

```bash
pytest tests/test_off_policy_is.py
```

Registered in the `cpu-unittest` matrix (`pr-test.yml.j2` after `test_policy_loss.py`).


EazyReal:
@zhuzilin could you review this one? It adds the off-policy IS hook at the policy-loss boundary using current-policy logprobs against rollout logprobs, preserving gradient flow through pi_theta while keeping the correction weight detached.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This adds an off-policy IS correction hook so training can account for rollout-vs-current policy mismatch when enabled. Could you review when you have bandwidth?

PR #2090, opened 2026-06-16
feat(loss): add pg_loss aggregation modes

EazyReal:
On main, slime has two built-in `pg_loss` normalizations: the default sample/rollout mean and the legacy `--calculate-per-token-loss` path. Prompt-group normalization and fixed-divisor normalization both require a custom `pg_loss` reducer today. That is fragile because the reducer has to compose correctly with CP slicing, micro-batch packing, and Megatron's final train-step divisor; a wrong constant factor silently changes the effective learning rate.

This PR moves the common `pg_loss` aggregation choices behind `--loss-aggregation`. For one train step, let `N` be the number of rollouts, `G = n_samples_per_prompt`, `P = N / G`, `M_i` be the valid-token count for rollout `i`, and `L = --loss-aggregation-divisor`.

| mode | step scalar | behavior |
| --- | --- | --- |
| `sample_mean` | `(1 / N) * sum_i token_mean(i)` | Keeps the current default behavior. |
| `token_mean` | `sum_i,t loss_it * mask_it / sum_i M_i` | Uses the existing `--calculate-per-token-loss` path under the new unified knob. |
| `prompt_mean` | `(1 / P) * sum_g token_mean(prompt_group_g)` | Groups rollouts by `Sample.group_index` and gives each prompt group one unit of weight. |
| `constant` | `sum_i,t loss_it * mask_it / (L * N)` | Uses a fixed token-scale divisor before the usual step average. |

The implementation reuses slime's existing reducer path instead of adding a parallel loss stack. Rollout builds `prompt_mask_sums` only for `prompt_mean`; `cp_utils.get_sum_of_sample_mean` still owns CP-aware summation; and the train step keeps its existing final scaling. For `prompt_mean`, the reducer scales the prompt-group sum by `n_samples_per_prompt`, so users get the requested prompt mean directly instead of a constant-offset objective that would need learning-rate compensation.

Compared with the other OSS designs checked for this change, this follows slime's distributed constraints rather than copying a central reducer wholesale. verl's actor reducer receives global denominators directly; SkyRL implements the scalar by pre-scaling advantages; prime-rl exposes only the global token mean; and AReaL can reuse its engine-level `local scalar + loss weight` contract. slime's Megatron/CP path has a different invariant: CP ranks own token shards, while sample, prompt, and constant denominators are whole-step data and the final train-step divisor already has response-count semantics. Keeping the mode selection inside the existing CP-aware reducer preserves those packing and train-step scaling rules.

Startup validation rejects configurations that would silently use the wrong denominator, such as `--loss-aggregation-divisor` outside `constant` or `--calculate-per-token-loss` together with `prompt_mean`/`constant`. The custom `pg_loss` reducer hook remains available and keeps precedence for non-standard objectives.

The focused tests cover the final `prompt_mean` scalar, CP/sample-denominator invariance, argument validation, and rollout metadata construction. The customization guide documents the built-in modes in both English and Chinese.

Tested with:

- `uv run --no-project --with pytest --with torch --with numpy --with httpx --with pyyaml python -m pytest -q tests/test_cp_utils.py tests/test_megatron_argument_validation.py tests/test_rollout_validation.py`
- `uv run --no-project --with ruff ruff check --config pyproject.toml docs/en/get_started/customization.md slime/backends/megatron_utils/loss.py slime/ray/rollout.py slime/utils/arguments.py tests/test_cp_utils.py tests/test_megatron_argument_validation.py tests/test_rollout_validation.py`
- `uv run --no-project --with black black --check --line-length 119 slime/backends/megatron_utils/loss.py slime/utils/arguments.py`


EazyReal:
Closing as superseded rather than recut.

Current main already has sample-mean (default + `rollout_mask_sums`), token-mean (`--calculate-per-token-loss`), and Dr.GRPO via `--custom-pg-loss-reducer-function-path`. The remaining 2090 mode, `prompt_mean`, has no in-tree caller; recutting it would reintroduce `--loss-aggregation` (or equivalent), which this recut is not replaying.

PR #2095, opened 2026-06-17
fix(dist): preserve new_group options across reloadable group reload

EazyReal:
## What changed

`ReloadableProcessGroup` now captures the original positional args and kwargs passed to `torch.distributed.new_group` and replays them verbatim in `reload_process_groups()`, instead of recreating each group with only `old_new_group(ranks=..., backend="nccl")`.

The gloo-skip and rank-extraction logic in the `new_group` monkeypatch is factored into `_is_gloo_group()` and `_get_group_ranks()` helpers (behavior-preserving; also removes the operator-precedence ambiguity in the old `a and b or c and d` gloo check).

Adds a CPU regression test (`test_reload_process_groups_preserves_new_group_timeout`) that drives a create -> destroy -> reload cycle and asserts `timeout` and `pg_options` survive the reload, and registers the test file in the cpu-unittest CI list.

## Why

On offload/reload, `reload_process_groups()` recreated each NCCL group with only `ranks` and `backend="nccl"`, dropping the original `timeout`, `pg_options`, and any other construction options. A job that launches with a longer distributed timeout would create its initial NCCL groups correctly but silently fall back to PyTorch's default timeout after a reload, so long-running actor train/update collectives could trip the NCCL watchdog even though the operator configured a larger timeout. The reload lifecycle is owned by this wrapper, so the wrapper is the only place that can preserve and replay the original `new_group` contract.

Gloo and single-rank groups keep their existing bypass (returned directly, not wrapped/registered).

## Validation

CPU-only, no GPU required:

- `pytest tests/test_reloadable_process_group_memory_check.py -q` -- the new test fakes `dist.new_group`/`old_new_group`, runs new_group -> destroy_process_groups -> reload_process_groups, and asserts the reloaded inner group is recreated with the same `ranks`, `backend`, `timeout`, and `pg_options` objects. It fails against the pre-fix code (which hard-codes `ranks=..., backend="nccl"`).
- The test file is registered in `.github/workflows/pr-test.yml.j2`'s cpu-unittest job so it runs in CI.


EazyReal:
@zhuzilin could you review this one? Reloadable distributed groups were recreating with default new_group options after reload; this preserves the original group options so timeout/backend semantics survive reload.

EazyReal:
@zhuzilin refreshed on latest main and checks are green. This preserves new_group options across reloadable group reloads, which avoids silently dropping timeout/backend settings after recovery. Could you review when you have bandwidth?

EazyReal:
Closing as duplicate of #2208 — same contract (replay original `new_group` args/kwargs on reload). Glad it landed.

This PR had been open with that change since Jun 17 (refreshed and pinged Jul 7).

EazyReal:
@zhuzilin thanks for landing this in #2208 — that's the fuller WORLD-reload design, and closing this as a duplicate of the `new_group` args/kwargs replay is the right call.

I'd opened this on Jun 17 for the subgroup-timeout hole and pinged Jun 25 / Jul 7. Totally fine that it shipped as part of a larger change. What would have helped on the contributing side is a short note that #2208 covered it, so I could have closed this instead of leaving a conflicted leftover.

If you're open to it, I'd love a slightly closer loop on communication for community PRs — even a one-line "landed in #N / still wanted / not in scope" is enough to know how to spend the next cycle. Happy to keep sending slime bugfixes and to reshape the open ones into whatever is easiest to review.

PR #2097, opened 2026-06-17
Recover Gemma native tool calls

EazyReal:
## Summary
- Add a schema-gated fallback for Gemma-native trailing `call:tool{...}` text after normal parser paths fail.
- Recover hidden-reasoning pseudo-calls when the visible body is empty, converting them to structured tool calls before adapter correction handling.
- Cover native command values, `<|"..."|>` strings, JSON-object arguments, and unknown-call no-op cases.

## Root cause
Gemma can emit a native pseudo tool call as decoded reasoning text while OpenAI wire `content`, `tool_calls`, and `function_call` stay null. The existing parser delegated to SGLang and XML fallbacks only, so the adapter returned no structured tool call and downstream clients injected no-tool-call correction text.

## Testing
- `uv run --no-project --python 3.12 --with pytest --with aiohttp --with torch python tests/test_agent/test_adapters.py`
- `uv run --no-project --python 3.12 --with ruff ruff check slime/agent/parsing.py tests/test_agent/test_adapters.py`
- `uv run --no-project --python 3.12 --with black black --check slime/agent/parsing.py tests/test_agent/test_adapters.py`
- `git diff --check`

### DIFF
--- a/slime/agent/parsing.py
+++ b/slime/agent/parsing.py
@@ -8,7 +8,6 @@ import logging
 import re
 from typing import Any
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


EazyReal:
closing for internal gemma 4 training only, will reopen if suitable again.

PR #2177, opened 2026-07-05
fix(tests): pass required --update-weight-local-checkpoint-dir in ext…

MengsD:
fix: test_qwen3_4B_external_pd.py

af3d7fff ("Disk-level delta weight sync", #2089) made
`--update-weight-local-checkpoint-dir` "Required for --update-weight-mode=delta
--update-weight-transport=disk" in slime/utils/arguments.py, but did not update
test_qwen3_4B_external_pd, which still builds delta_args without it. The test
aborts at arg validation:

  ValueError: --update-weight-mode=delta requires --update-weight-local-checkpoint-dir

Add a dedicated tempdir for the rollout-host-local checkpoint and pass it, 
mirroring the existing delta_dir handling; clean it up alongside delta_dir

### DIFF
--- a/tests/test_qwen3_4B_external_pd.py
+++ b/tests/test_qwen3_4B_external_pd.py
@@ -255,6 +255,13 @@ def execute():
 
     delta_dir_cm = tempfile.TemporaryDirectory(prefix="slime_external_pd_delta_")
     delta_dir = delta_dir_cm.name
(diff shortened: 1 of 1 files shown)


zhuzilin:
Sorry that I missed the PR... The issue has been solved in the latest slime. Close for now.

PR #2193, opened 2026-07-10
fix(update_weight): preserve grouped MoE expert axis during GLU rechunk

LLMShark:
### Background

The GLU rechunking logic in `update_weight/common.py` splits `linear_fc1` tensors along `dim=0` after TP all-gather to restore the gate/up partition order.

This is correct for:

- 2D dense or per-expert `linear_fc1.weight` tensors.
- 1D `linear_fc1.bias` tensors.

It is unsafe for grouped MoE tensors such as:

- Fused 3D weights with shape `[num_experts, 2 * ffn, hidden]`.
- Fused 2D biases with shape `[num_experts, 2 * ffn]`.

For grouped tensors, `dim=0` is the expert axis. Applying `chunk(2, dim=0)` would split the experts instead of the gate/up partitions and corrupt the gathered weights.

### Minimal reproduction and before/after

The previous implementation rechunked every `linear_fc1` tensor along `dim=0`.

Consider two TP partitions, each with shape `[2, 4, 3]`, and `partition_dim=2`:

- Expected gathered grouped-MoE shape: `[2, 4, 6]`
- Previous result after unconditional GLU rechunk: `[1, 4, 12]`

The expert dimension was incorrectly reduced from `2` to `1`, while the last dimension was incorrectly expanded.

Grouped-MoE weights use the following layout:

```text
[num_experts, 2 * ffn, hidden]
```

For these tensors, `dim=0` is the expert axis rather than the GLU gate/up axis. Rechunking along `dim=0` therefore reorganizes the expert weights incorrectly before model-specific conversion.

After this PR:

- 3D grouped-MoE weights preserve the expert axis and produce the expected `[2, 4, 6]` shape.
- 2D grouped-MoE biases are preserved intact.
- Existing GLU ordering remains unchanged for 2D dense/per-expert weights and 1D biases.
- The synchronous and asynchronous gather paths use the same predicate and are both covered by tests.
### Changes

Updated `slime/backends/megatron_utils/update_weight/common.py`:

- Added a shared `_requires_glu_rechunk(name, tensor)` predicate.
- GLU rechunking is applied only to:
  - 2D `linear_fc1.weight` tensors.
  - 1D `linear_fc1.bias` tensors.
- Grouped MoE tensors, including fused 3D weights and fused 2D biases, are preserved intact for downstream model-specific conversion.
- Both synchronous `all_gather_param()` and asynchronous `all_gather_params_async()` use the same predicate.

This prevents the synchronous and asynchronous weight-update paths from diverging while preserving the existing dense and per-expert GLU behavior.

### Tests

Added `tests/test_grouped_moe_glu_rechunk.py`, covering:

- 2D `linear_fc1.weight` still requires GLU rechunking.
- 1D `linear_fc1.bias` still requires GLU rechunking.
- Fused 3D grouped-MoE `linear_fc1.weight` does not get rechunked.
- Fused 2D grouped-MoE `linear_fc1.bias` does not get rechunked.
- `linear_fc2` is unaffected by GLU rechunking.
- Synchronous all-gather preserves the grouped-MoE expert axis.
- Asynchronous all-gather preserves the grouped-MoE expert axis.
- The existing 2D GLU partition ordering remains unchanged.

The test is included in the CPU CI matrix.

### Validation

- `git diff --check` passed.
- Python syntax checks passed.
- CPU unit tests passed locally: 8 passed.
- GitHub PR Test passed, including `test_grouped_moe_glu_rechunk.py`.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -643,6 +643,10 @@ jobs:
             },
             {
               "num_gpus": 0,
+              "test_file": "test_grouped_moe_glu_rechunk.py"
+            },
+            {
+              "num_gpus": 0,
               "test_file": "utils/test_hf_checkpoint_saver.py"
             },
             {
--- a/.github/workflows/pr-test.yml.j2
+++ b/.github/workflows/pr-test.yml.j2
@@ -85,6 +85,7 @@
         {'test_file': 'test_placement_group.py', 'num_gpus': 0},
         {'test_file': 'test_external_sglang_engines.py', 'num_gpus': 0},
         {'test_file': 'test_empty_colocated_weight_bucket.py', 'num_gpus': 0},
+        {'test_file': 'test_grouped_moe_glu_rechunk.py', 'num_gpus': 0},
         {'test_file': 'utils/test_hf_checkpoint_saver.py', 'num_gpus': 0},
         {'test_file': 'plugin_contracts/test_plugin_rollout_contracts.py', 'num_gpus': 0},
         {'test_file': 'plugin_contracts/test_plugin_runtime_hook_contracts.py', 'num_gpus': 0},
--- a/slime/backends/megatron_utils/update_weight/common.py
+++ b/slime/backends/megatron_utils/update_weight/common.py
@@ -12,6 +12,11 @@ from slime.backends.megatron_utils.misc_utils import strip_param_name_prefix
 from slime.utils.types import ParamInfo
 
 
+def _requires_glu_rechunk(name: str, tensor: torch.Tensor) -> bool:
+    """Return whether a tensor uses the dense or per-expert GLU partition layout."""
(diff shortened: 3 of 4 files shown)


LLMShark:
Hi maintainers, just a gentle ping on this PR. This is a focused correctness fix that prevents grouped MoE tensors from incorrectly rechunking the expert axis. The CI checks are passing and both synchronous and asynchronous paths are covered by tests. I would appreciate a review when you have time. Thanks!

LLMShark:
@zhuzilin  Hi, sorry to ping you directly. Since you recently updated this weight-update path in #2220, could you please review this focused grouped-MoE correctness fix when convenient, or point me to the appropriate reviewer? CI is green and the PR is mergeable. There is also a related post-load lifecycle fix in #2192, which has been updated against the latest main and passes CI. Thanks!

LLMShark:
@huang3eng Please help review~

PR #2197, opened 2026-07-12
fix: treat temperature=0 as greedy selection, untempered logprobs

EazyReal:
## Problem

`get_responses` and `get_log_probs_and_entropy` divide logits by `rollout_temperature` whenever it is not 1.0. Greedy rollout uses `temperature=0`, so that is a divide-by-zero.

SGLang treats T=0 as greedy token selection, but still returns finite untempered model log-probs. slime stores those as `rollout_log_probs` and builds PPO ratios from `exp(log π_θ − log π_old)`. Forcing greedy log-probs to 0 would make an on-policy sample start at `r = π_θ(a)`.

Value heads already pass `apply_temperature=False`.

## Change

`apply_rollout_temperature` is the shared rule at both logit sites:

- T in {0, 1}: leave logits unchanged
- otherwise: `logits / T`

T=0 also skips top-p nucleus replay. SGLang remaps T=0 to `top_k=1` and reports unmasked untempered log-probs.

## Test plan

- [x] `tests/test_rollout_temperature.py` — recomputed logprobs match `log_softmax` at T=0/1 and `log_softmax(logits/T)` at T=0.5; `get_responses` does not scale at 0; greedy skips top-p replay
- [x] `tests/test_value_temperature.py` — value head still ignores rollout temperature


EazyReal:
reviewer @zhuzilin commented offline that rollout_temperature must be > 0 in RL to ensure meaningful exploration, so we do not need to consider this case. 
opened and merged #2297 to ensure that.

PR #2205, opened 2026-07-14
perf: vectorize REINFORCE++ discounted returns

### LINKED ISSUE #2229: [Performance] Vectorize REINFORCE++ discounted-return computation
## Current limitation

`get_reinforce_plus_plus_returns` computes each token return in a Python reverse loop for every sample. For batched long responses, this host-side loop adds avoidable overhead. The existing chunked discounted scan used by GAE provides a reusable path for the same recurrence.

## Desired outcome

Compute REINFORCE++ discounted returns through the chunked/vectorized path while preserving numerical results, variable-length trimming, dtype and device, and context-parallel slicing behavior.

The related pull request reports the following CPU-only microbenchmark for return computation:

| Shape | Python reverse loop | Chunked scan | Speedup |
|---|---:|---:|---:|
| B=1, T=1,024 | 29.688 ms | 1.357 ms | 21.87x |
| B=8, T=4,096 | 1,778.590 ms | 4.979 ms | 357.21x |

These figures are not an end-to-end training-throughput claim.

## Related pull request

- #2205 — https://github.com/THUDM/slime/pull/2205

morluto:
## Problem

`get_reinforce_plus_plus_returns` computes every token return in a Python reverse loop for every sample. That host-side loop becomes expensive for batched long responses, while `ppo_utils.py` already contains a chunked discounted scan for GAE.

## Change

- Extract the existing recurrence into `chunked_discounted_returns`.
- Right-pad variable-length REINFORCE++ token rewards, scan the batch once, and trim each row before returning or context-parallel slicing.
- Reuse the same helper from `chunked_gae`.
- Register CPU tests for serial equivalence, chunk boundaries, discount edge cases, variable-length padding, dtype/device preservation, the REINFORCE++ caller, and GAE.

The change does not add a dependency or alter context-parallel gather/slice behavior.

## Correctness

The CPU tests compare the scan against the previous reverse recurrence for `float32` and `float64`, discounts `0`, `0.5`, `0.99`, and `1`, and lengths around the 128-token chunk boundary. Right-padded positions remain zero and are trimmed before caller-visible results.

```text
PYTHONPATH=. python3 tests/test_discounted_returns.py
45 passed in 2.36s
```

`git diff --check origin/main` also passes.

## CPU microbenchmark

Environment: 4-core DO-Premium-AMD CPU, Python 3.12.3, PyTorch 2.11.0, one PyTorch thread, random `float32` rewards, discount `0.99`, chunk size 128, one warmup, median of three repetitions.

The candidate used `chunked_gae(rewards, zeros_like(rewards), 0.99, 1.0)[0]`, which is the pre-existing recurrence extracted by this PR.

| Shape | Python reverse loop | Chunked scan | Speedup |
|---|---:|---:|---:|
| B=1, T=1,024 | 29.688 ms | 1.357 ms | 21.87x |
| B=8, T=4,096 | 1,778.590 ms | 4.979 ms | 357.21x |

This measures CPU return computation only. It is not a GPU or end-to-end training-throughput claim.

## Review notes

Suggested order: review `chunked_discounted_returns`, then the REINFORCE++ batching/trim path, then the CPU regression tests and generated CI entry.

## Disclosure

This contribution was prepared with AI-assisted research and implementation, followed by CPU validation and structured code review.

## Tracking issue

Fixes #2229

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -603,6 +603,10 @@ jobs:
             },
             {
               "num_gpus": 0,
+              "test_file": "test_discounted_returns.py"
+            },
+            {
+              "num_gpus": 0,
               "test_file": "test_rm_f1.py"
             },
             {
--- a/.github/workflows/pr-test.yml.j2
+++ b/.github/workflows/pr-test.yml.j2
@@ -75,6 +75,7 @@
         {'test_file': 'test_value_temperature.py', 'num_gpus': 0},
         {'test_file': 'test_cispo_loss.py', 'num_gpus': 0},
         {'test_file': 'test_ppo_logprob_entropy.py', 'num_gpus': 0},
+        {'test_file': 'test_discounted_returns.py', 'num_gpus': 0},
         {'test_file': 'test_rm_f1.py', 'num_gpus': 0},
         {'test_file': 'test_rm_gpqa.py', 'num_gpus': 0},
         {'test_file': 'test_rm_math.py', 'num_gpus': 0},
--- a/slime/utils/ppo_utils.py
+++ b/slime/utils/ppo_utils.py
@@ -397,7 +397,7 @@ def get_reinforce_plus_plus_returns(
 
     cp_size = mpu.get_context_parallel_world_size()
 
-    final_returns_chunks = []
+    token_level_rewards = []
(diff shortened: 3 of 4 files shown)


morluto:
great taste in anime @zhuzilin 
image

PR #2273, opened 2026-08-15
[ROCm] Support the INT4 QAT kernel on ROCm

LZ-QWQ:
The `int4_qat` kernel does not build on a ROCm PyTorch install, so the INT4 QAT path is unusable on AMD GPUs. This fixes the two blockers (a HIP warp-shuffle mask and a `const_data_ptr` symbol mangling mismatch under hipcc).

**Build**

- CUDA: unchanged. Every modification is behind `#if defined(__HIP_PLATFORM_AMD__)`  or `torch.version.hip is not None`,  so the CUDA compile path is byte-for-byte  identical.
- ROCm: `pip install .` just works. The gfx target comes from `PYTORCH_ROCM_ARCH`  (defaults to `gfx950`), since hipcc does not accept the nvcc-only flags.

**Verification** (MI355X / ROCm 7.2, CUDA reference from an RTX 4090)

- 144 cases covering the block shapes, tensor shapes and dtypes the kernel dispatches on: outputs are bit-exact against CUDA.
- E2E `tools/convert_hf_to_int4_direct.py` on Qwen3-0.6B: 196 layers, worst  relative error 0.1051.

### DIFF
--- a/slime/backends/megatron_utils/kernels/int4_qat/fake_int4_quant_cuda.cu
+++ b/slime/backends/megatron_utils/kernels/int4_qat/fake_int4_quant_cuda.cu
@@ -1,7 +1,15 @@
 #include <torch/extension.h>
 #include <ATen/cuda/CUDAContext.h>
 
+// HIP's __shfl_xor_sync requires a 64-bit mask (a wavefront is 64 lanes), so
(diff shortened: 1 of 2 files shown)


LZ-QWQ:
Closed by an accidental branch rename. Superseded by #2274 (same commit, no content change).

PR #2281, opened 2026-08-18
Fix Qwen3 loss masks for parallel tool results

Zeng-Weijun:
## Summary

Qwen3 chat templates group consecutive tool results into one rendered turn. The existing `qwen3` loss-mask path renders each message independently, which changes the token stream for parallel tool-call histories.

This change:

- uses the existing full-conversation, offset-based mask path whenever a Qwen3 conversation contains structured `tool_calls` or `tool` messages;
- preserves the current per-message Qwen3 behavior for conversations without structured tool history;
- shares the exact renderer with the Qwen3.5 path;
- adds a regression covering two parallel calls followed by consecutive tool results.

## Real tokenizer reproduction

Model: `Qwen3-30B-A3B-Instruct-2507`  
Base: `acac66163804b141b66272bcf2ae3b55f8edc643`

Full-template tokens vs. loss-mask-path tokens:

| Conversation | Base | This PR |
| --- | ---: | ---: |
| Natural turn | 27 = 27 | 27 = 27 |
| One tool call/result | 203 = 203 | 203 = 203 |
| Two parallel calls/results | **232 != 236** | **232 = 232** |

For all three rows on this branch, the token ID sequences are exactly equal and the mask length equals the token sequence length.

## Validation

- All four tests in `tests/utils/test_loss_mask_type_qwen35.py` passed under the project tokenizer runtime.
- The real Qwen3 tokenizer probe above passed, including exact ID equality for the parallel-tool row.
- `ruff check`, Python compilation, and `git diff --check` passed.

### DIFF
--- a/slime/utils/mask_utils.py
+++ b/slime/utils/mask_utils.py
@@ -33,9 +33,7 @@ class MultiTurnLossMaskGenerator:
             {"role": "user", "content": test_string},
         ]
         raw_token_ids = self.tokenizer(test_string, add_special_tokens=False)["input_ids"]
-        chat_template_token = self.tokenizer.apply_chat_template(
-            test_messages, add_special_tokens=False, tokenize=False
-        )
+        chat_template_token = self.tokenizer.apply_chat_template(test_messages, add_special_tokens=False, tokenize=False)
         chat_template_token_ids = self.tokenizer(chat_template_token, add_special_tokens=False)["input_ids"]
         idx_1, idx_2 = self.find_all_sublist_indices(chat_template_token_ids, raw_token_ids)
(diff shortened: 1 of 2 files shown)


Zeng-Weijun:
Closing as superseded by #2264, which landed after this branch was cut and fixes the same consecutive-tool-response tokenization path.

I reran the original real-tokenizer reproduction on upstream `main@41014d1f` with `Qwen3-30B-A3B-Instruct-2507`:

- natural: 27 = 27
- one tool call/result: 203 = 203
- two parallel calls/results: 232 = 232
- all three full-template and loss-mask-path token ID sequences are exactly equal

No additional upstream change is needed.

PR #2302, opened 2026-08-21
fix(rollout): normalize fan-out rewards by trajectory

sxl1993:
## Summary
- count each unique rollout once in the custom fan-out GRPO baseline
- broadcast the normalized trajectory reward to all emitted samples
- reject inconsistent sibling rewards and handle singleton groups without NaNs

## Motivation
For rewards [1, 1, 1, 0], where the first three samples belong to one rollout, the old hook used a sample-weighted baseline of 0.75. This change computes the trajectory-level baseline from [1, 0] as 0.5 and broadcasts [0.5, 0.5, 0.5, -0.5].

The default reward post-processing path is unchanged.

## Test
- added a CPU regression case to tests/test_sample.py


sxl1993:
@zhuzilin Please review #2302. It fixes reward normalization for fan-out GRPO by normalizing per trajectory. The main thing to check is the rollout_id grouping logic.

Issue #1111, opened 2025-12-15
Add cache to process inputs


nanjiangwill:
We currently apply `apply-chat-template` and process inputs in **two places**:

- **Data level**  
  https://github.com/THUDM/slime/blob/ac5328b0108e0c517e70b9369c33b237266a7e07/slime/utils/data.py#L58

- **Rollout level**  
  https://github.com/THUDM/slime/blob/ac5328b0108e0c517e70b9369c33b237266a7e07/slime/rollout/sglang_rollout.py#L99

This causes duplicated computation and unnecessary overhead.

### Why Two Stages Are Necessary
We cannot merge these into a single stage(right now, open to optimize):

- **Data level**: required to check context length early (with tokenization).
- **Rollout level**: performs tokenization and handles multimodal inputs.

### Proposed Solution
Introduce caching to avoid duplicate work.
1. Add a cache at the **data level** for processed chat templates.
2. At the **rollout level**, reuse the cached result when available.

liuhao-labs:
Hi, I rechecked this against current `main`. The duplicate work is still present: `filter_long_prompt()` tokenizes retained text prompts only to test their length, discards those IDs, and `_prepare_prompt_ids()` encodes the same rendered prompt again during the default rollout. #1113 improved tokenization correctness but did not retain or reuse those IDs.

I can submit a focused CPU-tested fix that caches only retained text prompt IDs when max-length filtering is enabled and consumes that cache in the default rollout path. I would intentionally leave multimodal processor outputs uncached: retaining pixel tensors for the whole `Dataset` and deepcopying them for every sampled sibling would create a much larger host-memory regression.

Would this text-only, default-rollout scope be the preferred first PR, or would you prefer a compact cache representation/API that custom rollout functions can explicitly opt into? I will keep `max_length=None`, custom rollout, and multimodal behavior unchanged.

nanjiangwill:
replied in https://github.com/THUDM/slime/pull/2309#issuecomment-5388361955

> Hi, I rechecked this against current `main`. The duplicate work is still present: `filter_long_prompt()` tokenizes retained text prompts only to test their length, discards those IDs, and `_prepare_prompt_ids()` encodes the same rendered prompt again during the default rollout. [#1113](https://github.com/THUDM/slime/pull/1113) improved tokenization correctness but did not retain or reuse those IDs.
> 
[...]

PR #2309, opened 2026-08-23
perf(rollout): reuse filtered prompt token ids

liuhao-labs:
## Summary

- retain token IDs already produced while filtering text prompts by max length
- store them as a compact transient uint32 payload and consume them once in the default SGLang rollout path
- keep custom rollout/generate/data-source contracts and multimodal processing unchanged

## Motivation

`filter_long_prompt()` currently tokenizes every retained text prompt to check its length, then `_prepare_prompt_ids()` encodes the same rendered prompt again before generation. This change reuses the first result for the stock train and eval paths.

The cache is intentionally limited to text prompts. Caching multimodal processor outputs would retain pixel tensors for the lifetime of the full Dataset and share mutable tensor storage across deep-copied siblings. The transient cache is excluded from `Sample.to_dict()` and fails closed to the existing tokenizer path.

Related to #1111.

## Validation

- pre-commit hooks: passed
- registered CPU test matrix: 51/51 files passed
- agent CPU suite: 64 passed, 1 skipped
- focused cache, serialization, eval, fully-async, and plugin contract coverage: passed
- 10,000-prompt local benchmark: second-pass `tokenizer.encode` calls reduced from 10,000 to 0; immediate total-time comparison improved from 2.65s to 0.57s on this machine

### DIFF
--- a/slime/rollout/data_source.py
+++ b/slime/rollout/data_source.py
@@ -14,6 +14,18 @@ from slime.utils.types import Sample
 logger = logging.getLogger(__name__)
 
 
+def _should_cache_prompt_token_ids(args) -> bool:
+    return (
+        args.rollout_function_path == "slime.rollout.sglang_rollout.generate_rollout"
+        and args.custom_generate_function_path is None
+        and args.data_source_path
+        in {
+            "slime.rollout.data_source.RolloutDataSource",
+            "slime.rollout.data_source.RolloutDataSourceWithBuffer",
+        }
+    )
+
+
 class DataSource(abc.ABC):
     @abc.abstractmethod
     def get_samples(self, num_samples: int) -> list[list[Sample]]:
@@ -80,6 +92,7 @@ class RolloutDataSource(DataSource):
                 tool_key=args.tool_key,
                 apply_chat_template=args.apply_chat_template,
                 apply_chat_template_kwargs=args.apply_chat_template_kwargs,
+                cache_prompt_token_ids=_should_cache_prompt_token_ids(args),
                 seed=args.rollout_seed,
             )
             if self.args.rollout_shuffle:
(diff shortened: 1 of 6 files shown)


nanjiangwill:
i checked this before and decided not to add it because re-tokenization especially in text-only case wont add too much latency. theoretically yes we should have this.

liuhao-labs:
Thanks for the context. I measured the retained text-only path before opening this, because the benefit depends on dataset size and prompt length.

On the same local workload (10,000 retained prompts, 2,189,001 prompt token IDs, an in-memory ByteLevel BPE), the current path made one batched tokenization call during filtering plus 10,000 per-sample `encode()` calls during rollout preparation. With this change, the second count is zero.

The immediate same-process comparison was:

- current path: 2.652 s total (0.586 s filter + 2.066 s prepare)
- cached path: 0.571 s total (0.541 s filter + 0.030 s prepare)
- rollout preparation: 98.6% lower, total for the two stages: 78.5% lower

The retained cache payload was 8,756,004 bytes (8.35 MiB) as packed uint32 IDs. It is only created when max-length filtering has already tokenized the prompt, only for retained text prompts on the stock rollout path, and it is consumed before Ray transfer. Multimodal processor outputs and custom rollout/data-source paths remain uncached.

So this is not intended to add tokenization work when filtering is disabled; it only reuses work that has already happened. The exact speedup will vary with tokenizer and prompts, but the invariant improvement is one fewer encoding pass per retained text prompt.

liuhao-labs:
Rebased onto current `main` (`3778dbf`) and preserved the existing implementation unchanged. The stale baseline failure from the removed `test_rollout_validation.py` is no longer part of the new CI run. Local checks passed: `test_filter_long_prompt.py` (22), `test_sample.py` (20), `test_rollout_data_utils.py` (6), plus the repository-pinned format/lint checks. @nanjiangwill, this is ready for another look when convenient.

PR #2310, opened 2026-08-24
fix(megatron): compact actor logits before projection

### LINKED ISSUE #2253: [Bug] Training/log-prob forward materializes full-vocab fp32 logits over the entire packed sequence → OOM for large-vocab, long multi-turn RL
### Bug Description

In the actor forward used for the policy loss and log-prob recompute, the model is called with labels=None, so GPTModel returns the full [1, T, vocab] logits and Float16Module upcasts them to fp32;
  get_log_probs_and_entropy then computes over the whole [T, vocab] tensor and only slices out the response spans afterward. For RL where only loss_mask==1 (model-generated) tokens contribute to the
  loss, this means logits are computed and upcast for every prompt and tool-output position too, then discarded. Peak memory therefore scales with the total packed sequence length × vocab, not with the
  number of response tokens. --log-probs-chunk-size doesn't help the peak, because it only chunks the loss-side reduction — the full fp32 [T, vocab] tensor is already materialized in the model forward
  before the loss runs.

  This becomes fatal with a large vocabulary and long multi-turn/agentic episodes. Concretely, with vocab ≈ 248k (Qwen3.5) and context 32k–64k, a single microbatch's fp32 logits alone can be tens of GB
  (e.g., a ~11.6k-token microbatch = 11.6k × 248320 × 4B ≈ 10.8 GB; near-context episodes are far larger), causing OOM in the training forward even with optimizer CPU offload. Note the tool-output/prompt
  tokens (which dominate long agentic sequences) are exactly the non-trainable positions being wasted.

### Steps to Reproduce

 1. Megatron backend, large-vocab model (Qwen3.5-9B, vocab_size=248320), TP=4/DP=2, colocate, on H100 80 GB GPUs.
  2. Long packed sequences — multi-turn/agentic rollouts where prompt + tool outputs dominate the sequence and only response tokens carry loss_mask=1; context length 32768–65536.
  3. Run a normal GRPO step (policy loss forward, or the old-actor/ref log-prob recompute), labels=None.
  4. Hit a microbatch whose packed length is large (e.g. ~11k+ tokens); observe the OOM in the training forward at the fp32 logit conversion. It reproduces even with --log-probs-chunk-size set and
  optimizer CPU offload enabled.

### Expected Behavior

Peak logit memory scales with the number of loss/response tokens (or is chunked), so large-vocab + long-context RL trains without OOM — since only response positions affect the loss.

### Actual Behavior

The full [1, T, vocab] fp32 logits for the entire packed sequence are materialized before any loss-side chunking, so peak ≈ T × vocab × 4 bytes regardless of how few tokens are trained. Example: a
  single ~11.6k-token microbatch = 11.6k × 248320 × 4B ≈ 10.8 GB just for that tensor (episodes near the context limit are far larger), causing torch.OutOfMemoryError in the forward (float16_to_fp32)
  even with optimizer offload. --log-probs-chunk-size doesn't help because it's downstream of the materialization.

### Environment

- slime version: 0.3.0 
- Python version: 3.12.7
- PyTorch version: 2.11.0+cu130
- CUDA/ROCm version: CUDA 13.0 (torch); toolkit 13.1; driver 580.82.07
- GPU type and count: 8× NVIDIA H100 (80 GB)
- OS: Alibaba Cloud Linux 3 (kernel 5.10.134-18.al8.x86_64)
- SGLang version (if relevant): 0.5.13
- Megatron-LM version (if relevant): megatron-core 0.16.0rc0

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

liuhao-labs:
## Summary

- Add an opt-in `--compact-actor-logits` path for built-in actor policy/SFT runs.
- Instantiate a Slime-owned `CompactLogitsGPTModel` only for eligible stock actor post-process chunks; critics, custom providers, non-final pipeline chunks, disabled configurations, and unsupported paths keep the original `GPTModel` behavior.
- Select `full_loss_masks` rows at the model post-process boundary, before full-vocabulary projection and fp32 conversion.
- Cover normal forward and combined 1F1B through normal instance-method dispatch, without modifying the process-global Megatron class or restoring the `megatron_patch` loader removed by #2316.
- Support compact log-probability/entropy computation with chunking, top-p replay, CP row mapping, empty masks, and autograd-safe scattering back to the existing response layout.
- Fail closed for MTP, config logging, custom model/loss/hooks, TIS/mismatch paths, debug capture, rollout log-probs, PPO reward-KL, and unsupported Megatron/output-layer configurations.
- Leave the optimization disabled by default; the Qwen3.6 coding-agent example opts in explicitly.

This avoids the full `[T, V]` projection/upcast peak described in #2253 while preserving the existing full-logit path wherever masked output values are not proven sufficient.

## Tests

- `80 passed` across compact model dispatch/provider ownership, compact loss/top-p/empty-mask behavior, policy loss, vocab-parallel log-probability/entropy, and CP invariance tests.
- New CPU test runs directly under the repository CI convention and is registered in both the workflow template and generated workflow.
- Ruff, Black, isort, compileall, workflow regeneration, YAML validation, shell syntax, and `git diff --check` passed.
- Added a real TP=2/NCCL parity test for sequence parallel off/on, selected logits, hidden gradients, output-weight gradients, and state restoration. The local host cannot run this CUDA/NCCL test, so it still requires the self-hosted changed-test job.
- The combined-schedule unit test verifies dynamic dispatch through a schedule node; pinned Megatron source was also audited to confirm `PostProcessNode` resolves `self.gpt_model._postprocess` at execution time.

Fixes #2253

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -771,6 +771,10 @@ jobs:
             },
             {
               "num_gpus": 0,
+              "test_file": "test_megatron_compact_logits.py"
+            },
+            {
+              "num_gpus": 0,
               "test_file": "test_model_provider_freeze.py"
             },
             {
--- a/.github/workflows/pr-test.yml.j2
+++ b/.github/workflows/pr-test.yml.j2
@@ -118,6 +118,7 @@
         {'test_file': 'test_empty_colocated_weight_bucket.py', 'num_gpus': 0},
         {'test_file': 'test_expert_routing.py', 'num_gpus': 0},
         {'test_file': 'test_glm5_indexer_short_context.py', 'num_gpus': 0},
+        {'test_file': 'test_megatron_compact_logits.py', 'num_gpus': 0},
         {'test_file': 'test_model_provider_freeze.py', 'num_gpus': 0},
         {'test_file': 'test_glm52_layerwise_comparison.py', 'num_gpus': 0},
         {'test_file': 'test_layerwise_alignment.py', 'num_gpus': 0},
--- a/examples/coding_agent_rl/run_qwen36_35b_a3b_swe_8nodes.sh
+++ b/examples/coding_agent_rl/run_qwen36_35b_a3b_swe_8nodes.sh
@@ -144,9 +144,10 @@ PERF_ARGS=(
    --recompute-granularity full
    --recompute-method uniform
    --recompute-num-layers 1
-   # max-tokens-per-gpu is one CP rank's slice of MAX_CONTEXT_LEN; log-probs are
-   # chunked along T to avoid OOM on long single trajectories.
+   # max-tokens-per-gpu is one CP rank's slice of MAX_CONTEXT_LEN. Project only
+   # loss-masked rows, then chunk the remaining log-prob reduction work.
    --max-tokens-per-gpu $((MAX_CONTEXT_LEN / CP_SIZE))
+   --compact-actor-logits
    --log-probs-chunk-size 1024
    --use-dynamic-batch-size
 )
--- a/slime/backends/megatron_utils/compact_logits.py
+++ b/slime/backends/megatron_utils/compact_logits.py
@@ -1,0 +1,190 @@
+import inspect
+from contextlib import contextmanager
+from contextvars import ContextVar
+
+import torch
+from megatron.core.models.gpt import GPTModel
+from megatron.core.tensor_parallel import gather_from_sequence_parallel_region
+from megatron.core.tensor_parallel.layers import ColumnParallelLinear
+
+_COMPACT_ACTOR_LOGITS = ContextVar("slime_compact_actor_logits", default=False)
+
+try:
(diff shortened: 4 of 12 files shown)


liuhao-labs:
CPU/pre-commit CI is green (58 successful checks). Could a maintainer add `run-ci-changed` for the modified `tests/test_ppo_logprob_entropy_gpu.py`? It contains the new real TP=2/NCCL forward-and-gradient parity coverage for sequence parallel off/on, which this local single-device environment cannot execute.

liuhao-labs:
Rebased this PR onto the latest `main` (`3778dbf`) and regenerated the workflow from `.github/workflows/pr-test.yml.j2`. The new CI run is queued on head `1a2a046`.

Local validation is still green: the focused CPU suite passes (`80 passed`), the new test passes via the direct CI-style runner (`31 passed`), and Ruff/Black/isort/diff checks pass. This host cannot run the TP=2/NCCL coverage in `tests/test_ppo_logprob_entropy_gpu.py`; could a maintainer please add `run-ci-changed` again so the self-hosted changed-test job executes it?

Issue #2188, opened 2026-07-08
[Bug] Colocate weight update fails with torch_memory_saver/offload because PyTorch CUDA IPC _share_cuda_ raises cudaErrorInvalidValue


Winnie-Lian:
### Bug Description

I am running a Slime GRPO job with Megatron backend, SGLang rollout engines, `--colocate`, `--offload-train`, and `torch_memory_saver` preload enabled. The job consistently fails during the first colocated weight update, before any training step completes.

Although `--update-weight-transport nccl` is configured, colocated engines still go through `UpdateWeightFromTensor`, which serializes flattened CUDA tensors with `MultiprocessingSerializer` / PyTorch CUDA IPC. In this environment that path fails in `storage._share_cuda_()` with `CUDA error: invalid argument`.

### Steps to Reproduce

1. Run a single-node 8xH100 Slime GRPO job with:
   - Megatron backend
   - SGLang rollout engines
   - `--colocate`
   - `--offload-train`
   - `--update-weight-mode full`
   - `--update-weight-transport nccl`
   - `--rollout-num-gpus 8`
   - `--rollout-num-gpus-per-engine 1`
   - `--tensor-model-parallel-size 4`
   - `--megatron-to-hf-mode bridge`

2. Enable `torch_memory_saver` preload:
   - `LD_PRELOAD=<torch_memory_saver_hook_mode_preload_cu12.abi3.so>`
   - `TMS_INIT_ENABLE=1`
   - `TMS_INIT_ENABLE_CPU_BACKUP=1`

3. Start training.

4. Observe that SGLang engines start successfully and Megatron loads the checkpoint, but the first `actor_model.update_weights()` fails inside `UpdateWeightFromTensor`.

### Expected Behavior

Slime colocate should either:

- successfully update colocated SGLang engines with CUDA IPC under `torch_memory_saver`, or
- provide a supported non-disk fallback, such as local NCCL weight update for colocated engines.

In particular, if `--update-weight-transport nccl` is configured, it would be helpful to either use NCCL for colocated engines too or document that colocated engines still use PyTorch CUDA IPC.

### Actual Behavior

The first colocated weight update fails in PyTorch CUDA IPC serialization:

```text
ray.exceptions.RayTaskError(AcceleratorError): ray::MegatronTrainRayActor.update_weights()
  File ".../slime/backends/megatron_utils/update_weight/update_weight_from_tensor.py", line ...
[...]

PR #2311, opened 2026-08-24
fix(args): reject unsupported colocated offload sync

liuhao-labs:
## Summary

- Reject the known-incompatible Megatron configuration that combines colocated local rollout engines, training offload, full weight sync, and the NCCL transport.
- Run the validation after custom YAML overrides, so a dangerous `custom_config` cannot bypass it.
- Keep valid boundaries unchanged: no-offload, zero local rollout GPUs, `debug_train_only`, non-colocated runs, and full+disk sync.
- Point users to the two supported choices: full+disk with a shared `--update-weight-disk-dir`, or `--no-offload-train` when memory permits.

Colocated full sync currently uses `UpdateWeightFromTensor` and CUDA IPC even when `--update-weight-transport=nccl` is selected. torch_memory_saver-backed offload allocations are not compatible with that PyTorch CUDA IPC serialization path, so this turns a late `storage._share_cuda_()` failure into an actionable startup error.

This is the narrow fail-fast stage; it does not claim to add a new non-disk colocated transport.

## Tests

- `PYTHONPATH=. python tests/test_megatron_argument_validation.py -q` (`26 passed`)
- The regression matrix covers derived colocate defaults, PPO-forced offload, YAML overrides, explicit no-offload, zero rollout GPUs, debug-train-only, and full+disk.
- Ruff, Black, isort, autoflake, compileall, and `git diff --check` passed.

Refs #2188

### DIFF
--- a/slime/utils/arguments.py
+++ b/slime/utils/arguments.py
@@ -2041,6 +2041,23 @@ def slime_validate_args(args):
     if args.only_train_params_name_list and args.freeze_params_name_list:
         raise ValueError("You can only specify ONE of: --only-train-params-name-list, or --freeze-params-name-list.")
 
+    if (
+        args.train_backend == "megatron"
+        and args.colocate
+        and args.rollout_num_gpus > 0
+        and not args.debug_train_only
+        and args.offload_train
+        and args.update_weight_mode == "full"
+        and args.update_weight_transport == "nccl"
+    ):
+        raise ValueError(
+            "Colocated Megatron weight updates with --offload-train and "
+            "--update-weight-mode=full --update-weight-transport=nccl are unsupported: "
+            "colocated engines use CUDA IPC, which is incompatible with torch_memory_saver offload. "
+            "Use --update-weight-mode=full --update-weight-transport=disk with a shared "
+            "--update-weight-disk-dir, or disable training offload with --no-offload-train."
+        )
+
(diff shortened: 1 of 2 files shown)


liuhao-labs:
Rebased onto the latest `main`; focused tests and formatting/lint checks pass locally. This is ready for review.

liuhao-labs:
The refreshed CI is green for the focused `test_megatron_argument_validation.py` job and pre-commit. The only failure is an unrelated [`test_advantage_whiten_cp.py` job](https://github.com/THUDM/slime/actions/runs/33358779351/job/99385927983): 5/6 parameter sets passed, while one multiprocessing worker exited with `SIGABRT`. This PR only changes argument validation and its focused tests. I do not have upstream Actions admin permission to rerun failed jobs; could a maintainer please rerun this failed job?

liuhao-labs:
The previously failing `test_advantage_whiten_cp.py` job was an isolated multiprocessing/Gloo teardown failure: the same test passed on the other current-main PR runs and all 6 parameter sets pass locally. I refreshed the single commit without changing its tree or PR diff so the current head received a clean check rollup. The new full [PR Test run](https://github.com/THUDM/slime/actions/runs/33612426688) is now green (62 passed jobs, 7 conditional skips, 0 failures), including the formerly failing job; pre-commit is also green. This is ready for review.

Issue #2285, opened 2026-08-19
[Question/Performance] Full message history is re-tokenized on every coding-agent turn


zhuyijie88:
## Background

  The current coding-agent adapters receive OpenAI/Anthropic messages from
  Claude Code or Codex and render the full message history on every turn:

  https://github.com/THUDM/slime/blob/main/slime/agent/adapters/common.py

  ```python
  translated, tools_schema = self._translate(body)
  prompt_ids = _render_token_ids(
      translated,
      tokenizer,
      tools=tools_schema,
      add_generation_prompt=True,
  )
```

  _render_token_ids() calls tokenizer.apply_chat_template(..., tokenize=True)
  for the complete message list.

  I understand that the exact output_ids and rollout logprobs returned by
  SGLang are preserved by TrajectoryManager, and that token drift is handled by
  realign/fork/masking. Therefore this question is about CPU scalability rather
  than token provenance correctness.

  ## Concern

  For a long-horizon agent trajectory, every request normally contains the
  complete and growing message history.

  If the prompt lengths over N turns are:

  L1, L2, ..., LN

  the adapter performs approximately:

  L1 + L2 + ... + LN

  tokens of CPU tokenization work.

  When the context grows approximately linearly with the number of turns, this
  becomes close to O(N^2) cumulative tokenization work per trajectory.

  This may become significant for:

  - 128K to 1M-token coding-agent trajectories;
  - many concurrent agent sessions sharing one adapter;
  - tool observations containing large source files or compiler logs;
  - repeated tool schemas and system prompts;
  - SGLang prefix-cache hits that make GPU prefill cheaper while CPU still
    re-renders and re-tokenizes the full prompt;

  - auto-compaction, sub-agent and retry paths.

  The tokenization call is also synchronous in the aiohttp request path. A long
  tokenization operation may therefore block other sessions served by the same
  adapter event loop.

  ## Questions

  1. Is full-history re-tokenization on every coding-agent turn the intended
     design?

  2. Has its CPU time, event-loop blocking time or throughput been benchmarked
     for long contexts and concurrent sessions?

  3. Are there existing metrics for:
      - chat-template rendering latency;
      - tokenizer latency;
      - number of input tokens re-tokenized per turn;
      - adapter queueing latency;
      - tokenizer CPU utilization?

  4. Is there a plan to support session-scoped incremental tokenization or an
     append-only token checkpoint, similar to a session-server TITO path?

[...]

PR #2328, opened 2026-08-26
perf(agent): isolate prompt rendering from event loop

liuhao-labs:
## Summary

- keep full-history `apply_chat_template(..., tokenize=True)` as the prompt-ID source of truth
- run prompt rendering on a single-thread, adapter-owned executor with a hard bound on submitted work; excess callers wait outside the executor queue
- retain capacity until the underlying future physically completes, skip canceled queued renders, and prevent canceled renders from reaching SGLang or creating session/trajectory state
- expose bounded recent-window metrics at `GET /debug/adapter_metrics` for prompt length, render latency, queue wait, worker utilization, event-loop lag, cancellations, and discarded work, including P50/P95/P99 distributions
- add CPU coverage for prompt-ID identity, event-loop responsiveness, admission bounds, repeated-cancellation races, and cleanup lifecycle

## Scope

This is an event-loop isolation and observability improvement. It deliberately keeps full-history rendering as the conservative fallback and does not claim to remove the cumulative O(N^2) tokenization cost. It remains orthogonal to the Qwen-specific append-only work in #2287.

Related to #2285.

## Validation

- `python -m pytest -q tests/test_agent`: 75 passed, 1 skipped
- registered direct CPU test: 11 passed
- repeated cancellation while admission is blocked returns within a bounded timeout and preserves the next permit
- Python 3.10-compatible cleanup/admission cancellation path, with no `Task.cancelling()` dependency
- Ruff, Black, isort, compileall, workflow regeneration, YAML parsing, and `git diff --check` passed

### DIFF
--- a/slime/agent/adapters/tokenizer_worker.py
+++ b/slime/agent/adapters/tokenizer_worker.py
@@ -62,6 +62,7 @@ class TokenizerWorker:
         self._loop: asyncio.AbstractEventLoop | None = None
         self._slots: asyncio.BoundedSemaphore | None = None
         self._acquire_tasks: set[asyncio.Task] = set()
+        self._close_cancelled_acquires: set[asyncio.Task] = set()
         self._futures: set[Future] = set()
         self._closed = False
         self._close_task: asyncio.Task | None = None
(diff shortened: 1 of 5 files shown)


slime/agent/adapters/tokenizer_worker.py
```diff
@@ -0,0 +1,345 @@
...
+        self._request_queued()
+        acquire_task = asyncio.create_task(self._slots.acquire())
+        self._acquire_tasks.add(acquire_task)
+        try:
+            await asyncio.shield(acquire_task)
+        except asyncio.CancelledError:
+            close_cancelled = acquire_task in self._close_cancelled_acquires
+            if not acquire_task.done():
+                acquire_task.cancel()
+            while not acquire_task.done():
```
zhuyijie88: The cancellation path should never wait indefinitely. Make semaphore-permit ownership explicit:

  1. On client cancellation, mark the request cancelled.
  2. Cancel the acquisition task.
  3. Attach a completion callback that releases the permit only if acquisition later succeeds.
  4. Return/raise CancelledError immediately, rather than looping while swallowing additional cancellations.
  5. Keep worker shutdown cancellation distinguishable through a separate close marker, as the second commit does.

  Add a regression test that issues two cancellations and asserts the request finishes within a short timeout, then verifies a subsequent render can
  acquire the permit.

liuhao-labs: Addressed in bbaff65. Admission cancellation no longer waits in a loop: it cancels the acquisition task, registers `_release_cancelled_acquire()` as the completion callback, and returns the original cancellation immediately. The callback owns cleanup of the acquire/close markers and releases the permit only when acquisition actually completed successfully.

I also added `test_repeated_cancellation_does_not_wait_for_admission`: the first render holds the only permit, the second request is cancelled twice and must finish within a short timeout, then a third render verifies that admission capacity was preserved. The existing handoff-race and shutdown tests remain in place.

Local validation after rebasing onto current `main`: the direct registered test has 11 passes, and the full agent suite has 75 passes / 1 skip. The updated CI run is now starting.

liuhao-labs: I strengthened the regression in 256bb42 after an independent differential check. The test now substitutes a bounded semaphore whose acquisition task deliberately delays completion after cancellation. That makes the previous wait-and-swallow loop exceed the short timeout, while the callback implementation returns the client cancellation immediately; the test then releases cleanup and verifies the next render still acquires the permit. The full agent suite remains 75 passed / 1 skipped locally, and a fresh Python 3.10 CI run has started.

PR #2331, opened 2026-08-26
Order pipeline weight-update process groups across rollout engines

chengcuiping:
## Problem

In non-colocated full NCCL weight updates with pipeline-parallel training and multiple independent SGLang rollout engines, each PP source creates a custom `slime-pp_<rank>` group containing itself and every engine GPU.

The PP sources are Ray actors invoked concurrently. Each rollout-engine actor uses a blocking HTTP call to join a group and processes methods serially. If engine 0 starts PP0 while engine 1 starts PP1, each group waits for the request queued behind the other engine's blocking call. The existing rollout-engine lock only covers later broadcasts, so it cannot prevent this group-creation cycle.

## Fix

- Have all training ranks execute connect/disconnect phases in global PP-rank order.
- In phase `k`, only the DP=0/TP=0 source for PP rank `k` creates or destroys its device group.
- Use the existing all-training-rank Gloo group to propagate phase failure before advancing.
- Apply a finite group/Ray wait timeout inherited from the distributed timeout and clean partial groups/futures/blocked engine actors fail-stop.
- Clear destroyed local group state before reconnect and release the transfer lock in `finally`.

This adds synchronization only when engines connect or disconnect, not to the normal bucket update path.

## Tests

- The deterministic opposite-arrival regression fails on `a067ce6f` and passes with this change.
- Unit tests cover timeout cancellation/engine cleanup, transfer-lock release, and failed reconnect state.
- Four-A100 deterministic NCCL opposite-order test: 20/20 passed.
- NCCL create/update/destroy stress: 50/50 passed with source/engine checksum agreement.
- Controls: PP=1 with two independent engines 20/20; PP=2 with one TP=2 engine 20/20.
- A no-delay Ray arrival-order probe observed the dangerous opposite order in 8/200 rounds. This is the opportunity rate for that arrival ordering on the test host, not a measured online deadlock rate.
- Production Slime helper plus Ray/NCCL integration: 20 rounds and 40 broadcasts passed with matching checksums.
- Two real SGLang HTTP/model-runner servers plus two Slime PP source actors: both groups connected, both online updates succeeded, both engines advanced to weight version 2, and generation changed identically after the update.
- 27 related CPU tests and ruff, black, isort, compileall, and diff checks passed.

The patched production-helper connect times were 0.121 s for PP0 and 0.115 s for PP1, comparable to 0.113/0.119 s in an earlier 10-round integration run. No meaningful connect-time regression was observed in this test.

## Environment note

The full Slime launcher could not run in the available A100 environment because that environment's installed `sglang-kernel` package contained SM100 but not SM80 common operations. No dependency was changed for this PR. The target path was instead validated successfully with real A100-compatible SGLang HTTP/model runners, Slime PP source actors, NCCL group creation, online weight updates, and functional generation checks.

## Risk

Connect/disconnect now performs one small Gloo all-reduce per PP group. A genuine timeout terminates only the affected rollout-engine actors because a serial actor blocked inside SGLang cannot service a queued cleanup RPC; recovery must recreate those engines. The steady-state weight broadcast path is unchanged apart from ensuring its lock is released on exceptions.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -587,6 +587,10 @@ jobs:
             },
             {
               "num_gpus": 0,
+              "test_file": "test_weight_update_group_ordering.py"
+            },
+            {
+              "num_gpus": 0,
               "test_file": "utils/test_megatron_role_config.py"
             },
             {
--- a/.github/workflows/pr-test.yml.j2
+++ b/.github/workflows/pr-test.yml.j2
@@ -72,6 +72,7 @@
       'extra_pip_deps': 'transformers wandb',
       'tests': [
         {'test_file': 'test_megatron_argument_validation.py', 'num_gpus': 0},
+        {'test_file': 'test_weight_update_group_ordering.py', 'num_gpus': 0},
         {'test_file': 'utils/test_megatron_role_config.py', 'num_gpus': 0},
         {'test_file': 'test_deep_ep_tms_patch.py', 'num_gpus': 0},
         {'test_file': 'test_stateless_adam.py', 'num_gpus': 0},
--- a/slime/backends/megatron_utils/update_weight/update_weight_from_distributed.py
+++ b/slime/backends/megatron_utils/update_weight/update_weight_from_distributed.py
@@ -4,6 +4,7 @@ import socket
 import time
 from argparse import Namespace
 from collections.abc import Callable, Iterator, Mapping, Sequence
+from datetime import timedelta
 
 import ray
 import torch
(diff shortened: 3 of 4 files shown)


chengcuiping:
@zhuzilin, could you please take a look at this PR when you have time?

It fixes a reproducible deadlock in non-colocated pipeline-parallel weight updates with multiple rollout engines by ordering the process-group connect and disconnect phases globally by PP rank. The steady-state weight broadcast path is unchanged.

The branch is mergeable and CI is green. I also validated the fix on 4×A100 with deterministic opposite-arrival tests, NCCL stress tests, and real SGLang weight updates. I would especially appreciate your feedback on whether the ordering and failure-handling design looks appropriate. I’m happy to adjust or reduce the scope if needed. Thanks!

PR #2347, opened 2026-09-03
[infra] Launch multi-node Slime jobs on Iris

penfever:
Launch one Ray cluster across a gang-scheduled Iris GPU job. Rank zero starts
the Ray head and training driver; sibling replicas join as workers and exit only
after the matching head attempt succeeds.

Multi-node launches require a unique S3 or shared-filesystem rendezvous. The
runtime refreshes the rendezvous while nodes join and rejects state left by an
older task attempt. Single-node launch behavior remains unchanged.


penfever:
Closing in favor of the experiment fork PR; this was opened against upstream by repository inference.

